PR #5, opened 2026-05-04
feat: one-command 'make full-setup' + per-request access logs

steventohme:
## Summary

- **`make full-setup`**: one-command bootstrap of the local docker-compose stack — boots Postgres + migrate + server, polls `/health`, seeds a Weave Router key (`rk_...`) into `.weave-router-key`, and optionally wires Claude Code via `install/install.sh --local`. Variants: `PLATFORM=cc` to auto-wire Claude Code; `KEY=rk_... BASE_URL=...` to skip compose and just point Claude Code at an existing router. Existing `make setup` (`migrate-up + seed`) is preserved for the host-DB Option B flow.
- **Per-request access log**: new `observability.AccessLog()` middleware emits one INFO line per request with method, path, status, latency_ms, client_ip. Auth-rejection logs (`invalid bearer prefix`, `bearer token did not match`) bumped Debug → Info so 401s are visible at the default `LOG_LEVEL=info` instead of disappearing into the void.
- **`docker-compose.yml`**: server now sets `ROUTER_DEV_MODE=true` so local dev works without seeding (matches `install.sh --local`'s assumption). Override in `.env.local` if you want bearer auth on locally.
- **`cmd/seed`**: rewritten output — calls the `rk_...` token a "Weave Router key" with a clear note that it is *not* an Anthropic key; recommends `install.sh` over raw `export`s; manual Claude Code path uses `ANTHROPIC_AUTH_TOKEN` (Claude Code's documented bearer slot) to dodge the name collision with the router's upstream `ANTHROPIC_API_KEY` env.
- **README**: rewrote Option A around `make full-setup PLATFORM=cc`, added a callout distinguishing the two key types, and switched `export ANTHROPIC_API_KEY=...` → `echo "ANTHROPIC_API_KEY=..." >> .env.local` so docker compose actually picks it up.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/observability/... ./internal/server/middleware/...` passing
- [x] `docker compose up --build -d` boots, `/health` returns 200, `/v1/messages` returns 200 in dev mode without auth
- [x] Verified live access-log output: every request now prints `level=INFO msg="http request" method=... path=... status=... latency_ms=...`
- [x] Verified auth rejections now log: `level=INFO msg="Auth rejected: bearer token did not match an active key"`
- [ ] End-to-end smoke: `make down && make full-setup PLATFORM=cc` from a clean state with only `.env.local` containing `ANTHROPIC_API_KEY`


makosblade (approved):
🪨🚢

PR #30, opened 2026-05-07
feat(auth): bind external API key ciphertexts to (external_id, provider) via AES-GCM AAD

makosblade:
## Summary

Binds every BYOK provider key ciphertext to its `(external_id, provider)` via AES-GCM AAD. A stolen row can no longer be replayed against a different installation or served as a different provider.

- `Encryptor.{Encrypt,Decrypt}` now take `externalID, provider string` and forward them as AAD to Tink. Tink returns a tag-mismatch error on Decrypt if either differs from the encryption-time value.
- `ExternalAPIKeyRepo.GetForInstallation` derives the AAD locally from `row.ExternalID + row.Provider` at decrypt time. No schema change required.
- Internal `aadFor(externalID, provider)` helper produces `externalID + 0x00 + provider`. The Weave backend must produce identical bytes.
- New `crypto_test.go` covers: golden-byte AAD format, round-trip, decrypt failure on mismatched externalID, decrypt failure on mismatched provider, distinct ciphertexts for distinct AAD, NoOp passthrough.

## Coordination with Weave

The Weave-side AAD encryption lands in workweave/WorkWeave (PR #7155). The two helpers must produce identical bytes for identical inputs; both repos pin the same golden-byte assertion (`aadFor(\"test-org\", \"anthropic\") == \"test-org\\x00anthropic\"`) so drift fails CI on either side.

Sequencing: this PR merges first; weave then bumps the router gitlink and merges.

## Risk

Production is empty on the BYOK side (Weave hasn't shipped), so no backfill or migration is needed. If any test/staging rows exist with the old (nil-AAD) ciphertext, they'll fail to decrypt post-deploy and the existing batch-decrypt warning path drops them — customer just re-saves the key.

## Test plan

- [x] \`wv mr tc\` passes
- [x] \`wv mr t\` passes (all 7 new crypto tests pass; full suite green)
- [ ] Cloud Run rollout on workweave-prod-01 succeeds
- [ ] Manual smoke: connect router for a test org, upsert an Anthropic key via the Weave PR's UI, confirm BYOK request lands at Anthropic

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/auth/crypto.go
+++ b/internal/auth/crypto.go
@@ -11,9 +11,11 @@ import (
 )
 
 // Encryptor handles AES-256-GCM encryption and decryption using Google Tink.
+// AAD binds each ciphertext to its (externalID, provider) so a stolen row
+// can't be decrypted for a different installation or provider.
 type Encryptor interface {
-	Encrypt(plaintext []byte) (ciphertext []byte, err error)
-	Decrypt(ciphertext []byte) (plaintext []byte, err error)
+	Encrypt(plaintext []byte, externalID, provider string) (ciphertext []byte, err error)
+	Decrypt(ciphertext []byte, externalID, provider string) (plaintext []byte, err error)
 }
 
 type tinkEncryptor struct {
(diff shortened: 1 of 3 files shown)


makosblade:
* **#30** <a href="https://app.graphite.com/github/pr/workweave/router/30?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/30?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #32, opened 2026-05-07
feat(providers/google): native Gemini adapter for thought_signature round-trip

steventohme:
## Summary

- Adds `providers/google.NativeClient` hitting `/v1beta/models/{model}:generateContent` (+ `:streamGenerateContent?alt=sse`) so Gemini 3.x preview models' opaque `thought_signature` round-trips across multi-turn tool use. The OpenAI-compat surface drops it, which 400s turn 2 of any tool-use loop.
- Adds OpenAI/Anthropic ↔ Gemini wire-format translators in `internal/translate/` (envelope-based, matching the existing `emit_openai.go` / `emit_anthropic.go` pattern).
- `proxy.Service` splits `ProviderGoogle` off from the OpenAI/OpenRouter cases in both `ProxyMessages` and `ProxyOpenAIChatCompletion`. For Anthropic-inbound, the response chain is Gemini SSE → OpenAI SSE → Anthropic SSE.
- `cmd/router/main.go` wholesale-swaps to `NewNativeClient`; the OpenAI-compat `Client` stays in the package as the revert path (bump the gitlink in WorkWeave).

Why now: empirically verified turn-2 400s on the OpenAI-compat path; see `eval/SWEBENCH_AGENTIC_STATUS.md` in WorkWeave.

## Files

Created (7):
- `internal/translate/emit_gemini.go` — `RequestEnvelope.PrepareGemini` (OpenAI- or Anthropic-source) + `GeminiStreamHintHeader`.
- `internal/translate/gemini_response.go` — non-streaming Gemini → OpenAI / Anthropic / error.
- `internal/translate/gemini_stream.go` — `GeminiToOpenAISSETranslator`.
- `internal/translate/gemini_test.go` — load-bearing tests (thought_signature round-trip both directions, system→systemInstruction, tool_choice, reasoning_effort, finish_reason, streaming text + tool_call + error).
- `internal/providers/google/native_client.go` — HTTP adapter; `x-goog-api-key` auth with BYOK precedence; mirrors `httputil.NewTransport` / `StreamBody`.
- `internal/providers/google/native_client_test.go` — fake-server tests for URL composition, stream toggling, hint-header stripping, BYOK override.
- `internal/providers/google/native_integration_test.go` — `//go:build google_integration`; 2-turn tool-use round-trip + text-only + streaming against the real API.

Modified (2):
- `internal/proxy/service.go`
- `cmd/router/main.go`

## Test results

`go build ./...` clean. `go vet ./...` clean. `go test ./... -tags=no_onnx -count=1` — all 16 packages green, including 16 new gemini-specific tests.

Integration test against the real Gemini API (`GOOGLE_PROVIDER_API_KEY=... go test -tags=google_integration -run TestNative ./internal/providers/google/`):

```
=== RUN   TestNative_TwoTurnToolUseRoundTripsThoughtSignature
    native_integration_test.go:94: turn 1 thoughtSignature length: 376
    native_integration_test.go:134: turn 2 response: ... "The weather in San Francisco is 62°F and sunny." ... finishReason: STOP
--- PASS: TestNative_TwoTurnToolUseRoundTripsThoughtSignature (1.73s)
=== RUN   TestNative_GenerateContentTextOnly
--- PASS (0.95s)
=== RUN   TestNative_StreamGenerateContent
--- PASS (0.77s)
PASS
```

The 376-byte `thoughtSignature` from turn 1 is the load-bearing piece — turn 2 returns 200 with a synthesized text answer instead of the prior 400.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -tags=no_onnx -count=1`
- [x] `go test -tags=google_integration -run TestNative ./internal/providers/google/`
- [ ] Bump WorkWeave gitlink and run `eval.run_swebench_agentic` against staging — confirms no `thought_signature` 400s in the run log.

## Revert plan

Revert the gitlink bump in WorkWeave; the OpenAI-compat `google.Client` is still in the package and `main.go`'s switch is the only call site to swap back.

### DIFF
--- a/internal/providers/google/native_client.go
+++ b/internal/providers/google/native_client.go
@@ -1,0 +1,154 @@
+package google
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"workweave/router/internal/observability/otel"
+	"workweave/router/internal/providers"
+	"workweave/router/internal/providers/httputil"
+	"workweave/router/internal/proxy"
+	"workweave/router/internal/router"
+	"workweave/router/internal/translate"
+)
+
+// NativeBaseURL is Google's native Generative Language base URL. Native
+// endpoints sit under /v1beta/models/{model}:generateContent — distinct from
+// the OpenAI-compat surface (which lives under /v1beta/openai). NativeClient
+// composes the per-request URL.
+const NativeBaseURL = "https://generativelanguage.googleapis.com"
+
+// NativeClient is the providers.Client adapter for Google Gemini's native
+// REST surface. The native API both returns and accepts the opaque
+// thought_signature field that multi-turn tool use against Gemini 3.x preview
+// models requires; the OpenAI-compat surface (see Client) does not.
+//
+// Auth is via the x-goog-api-key request header. BYOK credentials on the
+// request context take precedence over the deployment-level key.
+type NativeClient struct {
+	apiKey  string
+	baseURL string
+	http    *http.Client
+}
+
+// NewNativeClient is pooled for sustained traffic to a single host. baseURL
+// defaults to NativeBaseURL when empty.
+func NewNativeClient(apiKey, baseURL string) *NativeClient {
+	if baseURL == "" {
+		baseURL = NativeBaseURL
+	}
+	return &NativeClient{
+		apiKey:  apiKey,
+		baseURL: baseURL,
+		http:    &http.Client{Transport: httputil.NewTransport(5*time.Second, 5*time.Second)},
+	}
+}
+
+// Proxy posts the prepared Gemini-native body to :generateContent (or
+// :streamGenerateContent?alt=sse when the inbound request was streaming) and
+// streams the response back via the provided ResponseWriter.
+//
+// The streaming/non-streaming choice is communicated by the caller: the
+// translate.GeminiToOpenAISSETranslator wrapping w expects Gemini SSE only
+// when the upstream Content-Type advertises text/event-stream. Callers select
+// streaming by setting "X-Stream" on prep.Headers; we strip that synthetic
+// header before forwarding.
+func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep providers.PreparedRequest, w http.ResponseWriter, r *http.Request) error {
+	stream := prep.Headers.Get(translate.GeminiStreamHintHeader) == "true"
+	prep.Headers.Del(translate.GeminiStreamHintHeader)
+
+	method := ":generateContent"
+	query := ""
+	if stream {
+		method = ":streamGenerateContent"
+		query = "?alt=sse"
+	}
+	url := c.baseURL + "/v1beta/models/" + decision.Model + method + query
+
+	upstream, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(prep.Body))
+	if err != nil {
+		return fmt.Errorf("build upstream request: %w", err)
+	}
+	upstream.Header.Set("Content-Type", "application/json")
+	c.applyAPIKey(ctx, upstream)
+	for k, vs := range prep.Headers {
+		upstream.Header[http.CanonicalHeaderKey(k)] = vs
+	}
+	if stream {
+		upstream.Header.Set("Accept", "text/event-stream")
+	}
+
+	t := otel.TimingFrom(ctx)
+	t.StampUpstreamRequest()
+	resp, err := c.http.Do(upstream)
+	if err != nil {
+		return fmt.Errorf("upstream call: %w", err)
+	}
+	defer resp.Body.Close()
+	t.StampUpstreamHeaders()
+
+	providers.CopyUpstreamHeaders(w, resp)
+	w.WriteHeader(resp.StatusCode)
+	return httputil.StreamBody(resp.Body, resp.StatusCode, w, t)
+}
+
+// Passthrough forwards an inbound non-routing request (e.g. /v1/models) to the
+// native surface unchanged. The native API exposes /v1beta/models for model
+// discovery; the inbound /v1/... prefix is rewritten to /v1beta/.
+func (c *NativeClient) Passthrough(ctx context.Context, prep providers.PreparedRequest, w http.ResponseWriter, r *http.Request) error {
+	suffix := r.URL.Path
+	if rest, ok := strings.CutPrefix(suffix, "/v1/"); ok {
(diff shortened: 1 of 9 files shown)


internal/translate/emit_gemini.go
kevinahn7, line 1: A `thought_signature` attached to a **text-only** assistant turn is dropped on the next request. On the response side (`gemini_response.go`), a text-only reply's signature is preserved either as `message.thought_signature` (OpenAI path, `extractGeminiParts`'s `leadingSig`) or as `block.thought_signature` on a `{type:"text"}` block (Anthropic path, `buildAnthropicBlocksFromGemini`). But on the request side, `openAIAssistantToGeminiParts` only reads `thought_signature` off `tool_calls`/`function`, never off the message itself, and the `case "text":` branch here in `anthropicAssistantToGeminiParts` never reads `block["thought_signature"]` either (only the `tool_use` branch does, a few lines below). Since Gemini 3.x can require the signature to continue reasoning across turns, any multi-turn conversation whose assistant turn is plain text (no tool call) will silently lose the signature and can 400 on the next request — which is exactly the failure this native adapter was built to avoid.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/translate/gemini_stream.go
kevinahn7, line 1: `Finalize()`'s fallback path (`if !t.finished { t.emitFinalChunk("stop", nil) }`) can fire before `t.started` was ever set — e.g. if the upstream SSE stream closes without ever sending a chunk containing `candidates.0.content.parts` (only usage/empty chunks, or an immediate safety block with no content). In that case `emitFirstChunk()` (which sends the mandatory OpenAI `delta:{role:"assistant",...}` chunk) is never called, and the client only sees a bare `finish_reason:"stop"` chunk with no preceding role delta — a malformed OpenAI stream.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/translate/emit_gemini.go
kevinahn7, line 1: `pullOpenAISystemAndContents` emits a separate `{role:"user", parts:[functionResponse]}` Gemini content entry for *each* OpenAI `tool` message, rather than grouping consecutive `tool` messages that answer the same assistant turn's parallel tool calls into one `user` content with multiple `functionResponse` parts. The Anthropic path handles this correctly because Anthropic already nests multiple `tool_result` blocks in a single user message; the OpenAI path doesn't recombine them, so a multi-tool-call turn round-trips as several consecutive same-role Gemini contents instead of one.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/providers/google/native_client.go
kevinahn7, line 1: `Proxy` and `Passthrough` both do `upstream.Header[http.CanonicalHeaderKey(k)] = vs` for every header in `prep.Headers` *after* setting `Content-Type` and calling `applyAPIKey`. That ordering means any header `PrepareGemini` (or a future caller) puts in `prep.Headers` under the canonical `Content-Type` or `X-Goog-Api-Key` key would silently override the auth/content-type set just above — currently harmless since only `GeminiStreamHintHeader` is set (and it's deleted before this loop), but it's a latent footgun since nothing in this method's contract prevents `prep.Headers` from colliding with those two.

---
_Generated by [Claude Code](https://claude.ai/code)_

kevinahn7 (commented):
Low-effort diff pass on the new Gemini native-adapter code (skipped test files). 4 findings, most severe first.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/translate/emit_gemini.go
kevinahn7: Confirmed valid — but only half was fixed by later work on `main`. The **Anthropic** path now attaches the signature to text blocks (`anthropicAssistantPartsGJSON`, the `case "text":` branch reads `block.thought_signature`). The **OpenAI** path still dropped it: `openAIAssistantPartsGJSON` built the text part via `geminiTextPart(text)` and only applied `thought_signature` to `functionCall` parts, so a text-only assistant turn carrying a message-level `thought_signature` (exactly what `writeOpenAIMessageFromGemini` emits for text-only replies) lost it.

Fixed in 6e6dd6e: the message-level `thought_signature` is now attached to the leading text part via a new `geminiTextPartWithSig` helper, mirroring the Anthropic path. Added `TestPrepareGemini_FromOpenAI_TextOnlyTurnRoundTripsThoughtSignature` to lock it in. (Since #32 is merged, the fix is on branch `claude/pr-32-code-review-g80qsf`.)

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/translate/gemini_stream.go
kevinahn7: No longer applicable on `main`. The `Finalize()` this flagged has since been rewritten: the streaming path no longer does the unconditional `if !t.finished { emitFinalChunk("stop", nil) }`. It now delegates to a `StreamLifecycle` state machine — on EOF it only emits a terminal/failure chunk when `t.lifecycle.OutputStarted()` is true, which is exactly the guard this comment was asking for (don't emit a bare `finish_reason` chunk before the initial `role:"assistant"` delta). Nothing to change; resolving.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/translate/emit_gemini.go
kevinahn7: Already fixed on `main`. The OpenAI→Gemini path now buffers each message into a `contentEntry` and runs the result through `collapseConsecutiveRoles(entries)` before emitting (same call the Anthropic path uses). Consecutive `tool` messages become adjacent `user`-role `functionResponse` entries and get merged into a single `user` content with multiple `functionResponse` parts, so a parallel-tool-call turn no longer round-trips as several consecutive same-role Gemini contents. Nothing to change; resolving.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/providers/google/native_client.go
kevinahn7: Still present on `main` (the ordering was unchanged), and as noted it's a latent footgun rather than a live bug — the only header ever placed in `prep.Headers` is `GeminiStreamHintHeader`, which is `Del`'d before the loop. Hardened anyway in 6e6dd6e: both `Proxy` and `Passthrough` now copy caller-supplied `prep.Headers` **first**, then set `Content-Type` / `x-goog-api-key` / `Accept`, so the reserved headers stay authoritative regardless of what a future caller stashes in `prep.Headers`. No behavior change today. (On branch `claude/pr-32-code-review-g80qsf` since #32 is merged.)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #33, opened 2026-05-07
fix(proxy): keep session-key pinning active under eval-override headers

steventohme:
## Summary

- Eval-override headers (`x-weave-cluster-version`, `x-weave-embed-last-user-message`) used to disable **both** the session_key-based pin tiers (1+2) and the legacy apiKeyID LRU (tier 3). That made multi-turn agentic eval unsafe: cluster scorer re-ran fresh on every turn → free provider switches mid-conversation → first non-Gemini tool_call permanently breaks any subsequent Gemini route (no thoughtSignature → 400).
- Split `bypassEval` from `bypassLegacySticky`. Tier 1/2 (session_key keyed; each eval prompt has a unique system+user pair → unique key) stays active. Tier 3 (apiKeyID LRU; the eval shares one key across 500 instances) stays bypassed.

## Why now

PR #32 (native Gemini adapter) is correct, but the agentic SWE-bench 500-run still 400'd after a few instances. Trajectory inspection: turns 1–8 routed to OpenRouter (deepseek + qwen), turn 9 routed to Gemini-3.x → 400 `Function call is missing a thought_signature in functionCall parts. Additional data, function call \`default_api:bash\`, position 2`. Cross-provider conversation, not an adapter bug.

## Tests

- Updated `TestService_SessionPin_EvalOverrideHeaderBypassesAllTiers` → `…KeepsSessionKeyPinning`. New assertions: with eval-override header set, pin store IS consulted, pinned model wins over the cluster scorer's fresh decision.
- Existing `TestService_ProxyMessages_StickyBypassedByEvalOverrideHeaders` still asserts tier-3 bypass behavior unchanged.

`go build ./... && go test ./... -tags=no_onnx -count=1` clean.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/proxy/... -count=1`
- [x] `go test ./... -tags=no_onnx -count=1`
- [ ] Bump WorkWeave gitlink, deploy staging, re-run `eval/run_swebench_agentic --workers 2 --limit 500`. Expect zero `thought_signature` 400s.

### DIFF
--- a/internal/proxy/service.go
+++ b/internal/proxy/service.go
@@ -292,7 +292,16 @@ func (s *Service) ProxyMessages(ctx context.Context, body []byte, w http.Respons
 	installationID, _ := ctx.Value(InstallationIDContextKey{}).(string)
 	clientID := ClientIdentityFrom(ctx)
 	bypassEval := hasEvalOverrideHeader(r)
-	bypassSticky := bypassEval
+	// Tier-1/2 (session-key-based) pinning stays enabled for eval traffic. The
+	// session_key is derived from the system text + first user message, so each
+	// eval prompt produces a unique key; pinning per-session is exactly what an
+	// agentic harness needs to keep multi-turn tool-use on a single provider
+	// (mid-conversation provider switches break Gemini, which rejects function
+	// calls without thoughtSignature emitted by other providers). Tier-3 (the
+	// legacy apiKeyID LRU) stays bypassed because the eval harness shares one
+	// apiKeyID across all 500 instances — without that bypass, the first
+	// decision would stick across unrelated instances.
+	bypassLegacySticky := bypassEval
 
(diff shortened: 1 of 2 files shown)


kevinahn7 (commented):
Automated review (low-effort, diff-only pass): no correctness bugs found. The rename (`bypassSticky` → `bypassLegacySticky`) and the change to `pinEligible` (dropping the eval bypass so tier-1/2 session-key pinning stays active while tier-3 legacy apiKeyID LRU is still bypassed) are applied consistently at all four call sites, and the updated test (`TestService_SessionPin_EvalOverrideHeaderKeepsSessionKeyPinning`) matches the new intended behavior.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #159, opened 2026-05-15
feat: add Codex CLI support — credential passthrough + --codex installer

steventohme:
## Summary

Add first-class Codex (OpenAI's CLI coding agent) support so it can route through Weave the same way Claude Code does — with a one-line installer and credential passthrough that uses the user's own OpenAI plan key.

## Router credential passthrough

- **`internal/providers/openai/client.go`** — factored auth into `setAuth()` with three tiers: (1) BYOK creds on the request ctx, (2) the deployment-level `OPENAI_API_KEY`, (3) **passthrough of the inbound `Authorization` header**. The Anthropic client has had this pattern for the Claude plan flow; OpenAI was missing tier 3. Codex's `Authorization: Bearer sk-proj-...` now reaches `api.openai.com` even when no BYOK or deployment key is configured.
- **`cmd/router/main.go`** — added `openaiPassthroughEligible`, paralleling `anthropicPassthroughEligible`. When selfhosted has no `OPENAI_API_KEY`, OpenAI is still added to `deploymentEligible` so the cluster scorer can route there (it would otherwise be invisible to argmax). Stays out of `envKeyedProviders` so hard-pin compaction won't land on a model that needs a credential the request might not carry.

## Installer (`install/install.sh`)

- New `--codex` flag (target = Codex CLI) and `--claude` flag (explicit target = Claude Code, for symmetry with `--codex` so pipelines can skip the picker).
- Without either flag, an interactive picker asks the user. Non-interactive runs (`--non-interactive`, CI, `curl | sh`) keep the prior Claude Code default so existing pipelines don't change semantics.
- `write_codex_config` writes a managed `[model_providers.weave]` block (with `model_provider = "weave"` and `wire_api = "chat"`) to `~/.codex/config.toml`, delimited by begin/end markers. Re-install replaces the block in place; top-level `model_provider =` outside the markers is stripped so TOML doesn't end up with a duplicate key. User content outside the markers is preserved.
- Banner adapts ("Codex installer" / "Claude Code installer"). jq is no longer required for the Codex path (TOML uses awk).

## Uninstaller (`install/uninstall.sh`)

- `--codex` strips the managed block from `~/.codex/config.toml` and removes a top-level `model_provider = "weave"` that lives outside the markers. Other top-level `model_provider` values are preserved so we don't silently yank the user back into the OpenAI default if they had their own.

## Seed command

- **`cmd/seed/main.go`** — prints Codex setup snippets alongside the existing Claude Code / Cursor blocks so `make seed` shows a one-shot Codex config too.

## Test plan

- [x] Build clean: `go build -tags no_onnx ./...`
- [x] `go test -tags no_onnx -short ./internal/providers/openai/... ./cmd/router/...` passes
- [x] `bash -n install/install.sh` and `bash -n install/uninstall.sh` syntax-clean
- [x] Codex install smoke test in temp dir produces valid TOML with the managed block
- [x] Re-install (key rotation) replaces the block in place (no duplicate `model_provider`)
- [x] Install over a pre-existing user config preserves the user's `[profiles.*]` / `[model_providers.*]` sections
- [x] Uninstall round-trip removes the managed block and preserves user content
- [ ] Interactive picker (needs real TTY — verify by hand with `bash install/install.sh --local`)


steventohme:
Pushed `ed19e8f` addressing all three Cursor-bot findings:

**1. Router key forwarding in OpenAI passthrough (HIGH security)** — `setAuth()` now strips `Authorization: Bearer rk_...` before forwarding. Mirrors the `!HasAPIKeyPrefix` guard in `proxy.ExtractClientCredentials`. Locked in by `TestSetAuth_PassthroughRejectsRouterKey` in `internal/providers/openai/setauth_internal_test.go`.

**2. Cross-provider credential leakage via passthrough (HIGH security)** — split passthrough eligibility out of `deploymentEligible` into a new `passthroughEligibleProviders` set on `proxy.Service`. In `enabledProvidersForRequest`, passthrough providers join the eligible set **only when the inbound surface matches**. An Anthropic-surface request can no longer route to OpenAI in passthrough mode (and vice versa) — env-keyed providers stay eligible cross-surface since their creds don't depend on inbound headers. This also closes the pre-existing symmetric hazard the bot flagged for `anthropicPassthroughEligible`. Tests: `TestEnabledProvidersForRequest_PassthroughIsSurfaceScoped` + `TestEnabledProvidersForRequest_DeploymentKeyedStillCrossSurface` in `internal/proxy/enabled_providers_internal_test.go`.

**3. Scope prompt shows Claude paths for Codex target (MEDIUM)** — the interactive scope prompt now branches on `$target`: when `--codex`, the option descriptions read `~/.codex/` / `<repo>/.codex/` and "run codex" instead of the Claude equivalents.

PR #313, opened 2026-06-04
fix(cluster): raise EmbedTimeout to 2.5s to kill cold-start embedder 503s

steventohme:
## Problem

The cluster scorer's *"router unavailable: cluster scorer failed and no fallback is configured"* 503s are **entirely embedder timeouts** — the in-process ONNX (hugot) embedder overruns its 1500ms `EmbedTimeout` with `context deadline exceeded`. **~196 in prod over 10 days**, clustered into cold-start bursts (dense sub-second runs right after a replica boots).

Root cause: ONNX Runtime allocates its memory arena + intra-op threads **lazily on the first inference**, so a cold replica's opening burst all blow the 1.5s budget until it warms. There was no boot warm-up.

## Why not a fallback

This is **not** a missing model-fallback. The router's no-fail-open design is deliberate and correct (a default-model shortcut silently masks routing regressions — see `cluster/CLAUDE.md`). The fix is **embedder reliability**, not degrading routing quality.

## Fix

- **Warm-up at boot:** `NewEmbedder` runs one inference before returning, moving ONNX's lazy-init cost off the request path. It also validates the pipeline end-to-end — a broken model now fails boot loudly instead of 503ing every request at runtime.
- **Timeout cushion:** raise default `EmbedTimeout` 1500ms → 2500ms for contention spikes (still overridable via `ROUTER_CLUSTER_EMBED_TIMEOUT_MS`).

## Verification

Diagnosis from prod logs: all these 503s log `embed failed … context deadline exceeded` at `embed_ms`≈1500, bursting right after replica boot. `go build`/`vet`/`gofmt` clean; cluster package tests green. (The ONNX path is exercised by the `onnx_integration`-tagged suite, which needs the model assets; warm-up doesn't change `Embed` semantics.)

### DIFF
--- a/internal/router/cluster/AGENTS.md
+++ b/internal/router/cluster/AGENTS.md
@@ -10,7 +10,7 @@ AvengersPro-derived primary router (arXiv 2508.12631, DAI 2025). **P0.** Full de
 
 Package compiles in **two layered modes via build tags**:
 
-- `embedder_onnx.go` vs `embedder_stub.go` — gated by `no_onnx`. Default builds compile the real hugot-backed embedder; `-tags=no_onnx` swaps in a stub `NewEmbedder` that always errors. Used by contributors without `libonnxruntime`.
+- `embedder_onnx.go` vs `embedder_stub.go` — gated by `no_onnx`. Default builds compile the real hugot-backed embedder; `-tags=no_onnx` swaps in a stub `NewEmbedder` that always errors. Used by contributors without `libonnxruntime`. `buildClusterScorer` warms the embedder at boot (5s cap) to burn ONNX's lazy graph-optimization cost. The cold-start / burst `EmbedTimeout` 503s seen in prod were timeout-bound (concurrent inferences just over the budget), so the lever is `EmbedTimeout`, default 2.5s (override: `ROUTER_CLUSTER_EMBED_TIMEOUT_MS`).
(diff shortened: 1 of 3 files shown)


steventohme:
Addressed the Bugbot finding (unbounded boot warm-up). Good catch — it surfaced that `buildClusterScorer` *already* warms the embedder at boot with a 5s cap, so the `NewEmbedder` warm-up I'd added was both redundant and unbounded. Reverted it.

The real gap was that the existing warm-up primed with a 7-char `"warmup"` string — ONNX does shape-specific allocation on the first inference at a given sequence length, so the first real (long) prompt on a cold replica still paid that cost. The existing capped warm-up now primes with a representative max-length input instead. Timeout bump kept.

steventohme:
Addressed both Bugbot findings on the max-length warm-up — by reverting it.

The "5s warmup cap" one is right: growing the warm-up input to MaxPromptChars while keeping the 5s cap risked a **boot panic** on a slow cold replica, trading runtime 503s for a startup failure. And on the "tail truncation" one — for a uniform synthetic string, prefix vs tail is shape-identical (ONNX allocation is sequence-length-driven, not content/position-driven), so that framing doesn't hold.

Stepping back on the data: every failure logs `embed_ms ≈ 1500` (exactly the cap), in concurrent bursts — i.e. **timeout-bound** (inferences landing just over budget), not first-inference graph-opt cold cost (which the existing tiny warm-up already burns and which would log multi-second `embed_ms`). So the max-length warm-up was marginal *and* risky.

Reverted `main.go`/`embedder_onnx.go` to baseline; the PR is now just the evidence-backed lever — `EmbedTimeout` 1500ms→2500ms (still env-tunable). Renamed accordingly.

PR #333, opened 2026-06-08
fix(translate): clamp over-length tool-call ids on the OpenAI emit path

steventohme:
## Problem

OpenAI rejects `tool_calls[].id` longer than 64 chars:

```
400: Invalid 'messages[3].tool_calls[0].id': string too long. Expected a string with maximum length 64, but got a string with length 1411 instead.
```

The Anthropic→OpenAI emit path (`buildOpenAIToolCall` / `buildOpenAIToolResultMessage`) passed the raw id straight through — no length clamp, and (unlike the Anthropic emit path) no char-sanitize either. Cross-format round-trips occasionally produce ids far over the limit.

Found in the v0.66 Pro bake-off upstream-error logs (1 occurrence, gpt-5.5). Sibling fix to #332 (the system-role hoist that covered 35/38 of those errors); this covers the over-length-id case.

## Fix

- Add a length clamp to `sanitizeToolUseID`: ids over 64 chars → deterministic `tc_<sha1hex>` (43 chars). Hashing the **original** id means a `tool_use` block and its matching `tool_result` (same original id) map to the same clamped value, so the pair stays linked.
- Apply `sanitizeToolUseID` at both OpenAI emit sites (`tool_calls[].id` and `tool_result.tool_call_id`).

Short ids (the common case) are unchanged.

## Tests

- `TestCrossFormat_AnthropicToOpenAI_OverlongToolUseIDClampedAndPaired` — 1411-char id → emitted id ≤64 and `tool_call_id` still equals it.
- Existing tool-conversation tests (short ids unchanged) stay green.

`go build ./...`, `go vet`, `gofmt`, full `internal/translate` suite green.


steventohme:
Good catch on the shared-helper regression — fixed in 4c6a7e2.

The length clamp was in `sanitizeToolUseID`, which the Anthropic and Gemini emit paths also use. Since `embedSignatureInID` smuggles Gemini `thoughtSignature`s into tool ids (making them intentionally >64 bytes), hashing there would have broken `extractSignatureFromID` and dropped Gemini 3.x tool history after a provider switch.

Fix: reverted `sanitizeToolUseID` to char-only, and added an **OpenAI-only** `clampOpenAIToolCallID` that (1) drops the OpenAI-irrelevant thought signature, (2) keeps the bare id when ≤64, (3) hashes only a genuinely over-length id. Added a regression test asserting the shared path preserves long thought-signature ids.

PR #342, opened 2026-06-09
feat(catalog): add claude-fable-5 (Mythos-class, $10/$50, native 1M context)

steventohme:
## What

Adds Anthropic's **Claude Fable 5** (`claude-fable-5`, GA 2026-06-09 — the first publicly available Mythos-class model) to the router as a catalog/passthrough model.

- **catalog.go** — `TierHigh`, $10/$50 per MTok, cache-read multiplier 0.10, `ContextWindow: 1_000_000`. Fable ships 1M context **by default** (no `context-1m-2025-08-07` beta header needed), unlike Opus 4.6+ where the catalog conservatively stays at 200K and the pre-filter expands on the beta header.
- **model.go** — `anthropicAdaptive` spec: adaptive thinking is always on (`thinking.type=disabled` is rejected upstream), same wire shape as Opus 4.8. `CapExtendedContext` makes the beta header a harmless no-op.
- **force_model.go** — `fable` / `fable-5` / `fable5` / `claude-fable` aliases for the `x-weave-force-model` header.
- **translate.go** — Fable's safety classifiers decline with the new `stop_reason: "refusal"` on HTTP 200; map it to OpenAI `finish_reason: "content_filter"` instead of falling through to `"stop"` so OpenAI-format clients can distinguish a refusal from a clean stop.
- **genprices regen** — price tables in `install/install.sh` + `install/cc-statusline.sh`; OpenCode + pi model lists get a Fable entry (1M context / 128k max output).
- **pricing_test.go** — lookup row for the new pricing.

## What this deliberately does NOT do

Fable is **not** added to the cluster roster (`model_registry.json` / `model_features.json`). It has no RouterArena/SWE-bench labels yet, so the no-op onboarding path (#340) would have to proxy Opus 4.8's quality column — identical quality at 2× the price means the scorer would never pick it. Roster onboarding should follow the normal lifecycle: gather bench labels → offline screen → live bake-off → promote.

Until then it's reachable via passthrough and `/force-model fable`.

## Notes for reviewers

- Fable's `fallbacks` API param (server-side refusal retry, beta) is not wired — client passthrough of the param works unmodified; native router-side fallback-to-opus on refusal could be a follow-up.
- Specs triple-checked against platform.claude.com models overview + the "Introducing Claude Fable 5 and Claude Mythos 5" launch notes.

## Testing

- `go build ./...` clean
- `go test ./...` all green (incl. new pricing row, catalog invariant tests)
- `go run ./cmd/genprices` output committed


steventohme:
Superseded by #347 (squashed onboarding PR including this catalog change + both candidates).

PR #350, opened 2026-06-09
feat(docker): pin Qwen3 embedder assets to weave-eng HF repo

aminsamir45:
## Summary

Final wiring step for the dual-embedder rollout (#346, #349). The Qwen3-0.6B ONNX export (INT8 per-channel, down_proj/o_proj fp32, last-token pooling baked in) is uploaded to the private [weave-eng/qwen3-embedding-0.6b-onnx-router](https://huggingface.co/weave-eng/qwen3-embedding-0.6b-onnx-router) repo.

- Pins `HF_QWEN_REPO=weave-eng/qwen3-embedding-0.6b-onnx-router`, `HF_QWEN_REVISION=a43740ed1bb6436deaa4b79c9a132a3d98756ace` (upload commit SHA).
- Uploaded `model.onnx` matches the export that passed the parity gates (int8 0.983 vs reference, Go-vs-Python 0.981) and is byte-identical to the fixture's source (1,042,047,545 bytes).
- Docs: note the private repo + `--secret id=hf_token` requirement.

## Deploy note

Image builds now require `docker build --secret id=hf_token,src=<token-file>` (the Dockerfile already threads the token through every HF curl). Verified the resolve URL returns 200 with token, 401 without. Jina-only deploys can pass `--build-arg HF_QWEN_REPO=` to skip.

## Test plan

- [ ] Deploy-pipeline build passes with the hf_token secret mounted
- [ ] Booted image constructs the Qwen embedder when a Qwen bundle is the default version

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,13 +24,14 @@ ARG HF_MODEL_REPO=jinaai/jina-embeddings-v2-base-code
 # silently picked up new weights" surprises. Bump deliberately.
 ARG HF_MODEL_REVISION=516f4baf13dec4ddddda8631e019b5737c8bc250
 # Second embedder: Qwen3-Embedding-0.6B exported with last-token
-# pooling baked into the graph (scripts/export_qwen3_onnx.py). Empty
-# HF_QWEN_REPO skips the pull — the runtime constructs embedders
-# lazily, so deploys serving only Jina bundles don't need the asset.
-# Once the export is uploaded, set the repo default here and pin
-# HF_QWEN_REVISION to the upload commit SHA printed by the script.
-ARG HF_QWEN_REPO=
-ARG HF_QWEN_REVISION=main
+# pooling baked into the graph (scripts/export_qwen3_onnx.py). The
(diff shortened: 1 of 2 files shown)


aminsamir45:
Update after checking where existing assets actually live:

- **Jina precedent is a public HF pull** pinned to a commit SHA in the Dockerfile; cluster artifacts (v0.X) are committed in git. Nothing router-related lives in GCS (`prod-01-models-bundle` is the estimation product).
- **The deploy build passes no HF token** (monorepo `deploy.yml` → depot build of the `router-internal/router` submodule), so a private repo default would 401 the next router image build.
- Flipped `weave-eng/qwen3-embedding-0.6b-onnx-router` to **public** (Apache-2.0 derivative, same model as Jina) — verified anonymous resolve works. Updated Dockerfile/docs wording accordingly; no deploy-pipeline changes needed anymore.
- Added a model card to the HF repo documenting export provenance and the pin-by-SHA policy.

PR #383, opened 2026-06-12
fix(translate): strip unsigned thinking blocks before Anthropic upstream

rohith500:
## Problem

OSS models (DeepSeek, Qwen via Fireworks/DeepInfra/Bedrock) return
`reasoning_content` in their responses. The router translates this into
Anthropic `thinking` content blocks — but without an Anthropic-issued
`signature` field.

When a subsequent turn routes to any Anthropic thinking-capable model,
the API rejects the request with:

  400: thinking blocks in messages must have a signature

This affects any mixed OSS+Anthropic session with thinking enabled —
which is a core routing use case.

## Root cause

`resolveAnthropicOverrides` in `envelope.go` only strips thinking blocks when:
1. The target model has no thinking capability at all, OR
2. `ModelSwitched` is explicitly set

Neither condition catches the cross-provider case where an OSS model
produced unsigned thinking blocks and the session then routes to Claude.

## Fix

- `historyHasUnsignedThinkingBlock()` — detects thinking blocks missing
  a `signature` field in conversation history
- `StripUnsignedThinkingBlocks` — new flag on `EmitOverrides`, more
  surgical than `StripThinkingBlocks`: removes ONLY unsigned blocks,
  leaving valid Anthropic-signed thinking blocks from earlier Claude
  turns intact
- `stripUnsignedThinkingBlocksBytes()` — walks `messages[*].content[*]`
  and removes only thinking blocks with no signature

## Tests added

- `TestAnthropicSameFormat_UnsignedThinkingStrippedForThinkingCapableModel`
  — unsigned block stripped, sibling text block survives
- `TestAnthropicSameFormat_SignedThinkingPreservedForThinkingCapableModel`
  — signed block preserved when no unsigned blocks present
- Updated `ThinkingBlocksKeptForCapableModel` fixture to include a valid
  signature (the unsigned fixture was itself a latent bug)

### DIFF
--- a/internal/translate/envelope.go
+++ b/internal/translate/envelope.go
@@ -358,6 +358,10 @@ type EmitOverrides struct {
 	DefaultMaxTokensValue   int64
 	InjectStreamUsage       bool
 	StripThinkingBlocks     bool
+	// StripUnsignedThinkingBlocks removes only thinking blocks that have no
+	// Anthropic-issued signature, leaving signed thinking blocks intact.
+	// Used when OSS-generated unsigned blocks are detected in history.
+	StripUnsignedThinkingBlocks bool
 	// SanitizeToolUseIDs rewrites tool_use.id / tool_use_id values that contain
 	// characters outside ^[a-zA-Z0-9_-]+$. Always set when targeting Anthropic:
 	// non-Anthropic upstreams (e.g. Kimi-k2.6) emit IDs like "functions.Read:0"
@@ -408,6 +412,12 @@ func applyOverrides(body []byte, ov EmitOverrides) ([]byte, error) {
 		out, err = stripToolUseThoughtSignatureBytes(out)
 		if err != nil {
 			return nil, fmt.Errorf("strip tool_use thought_signature: %w", err)
+		}
+	}
+	if ov.StripUnsignedThinkingBlocks {
+		out, err = stripUnsignedThinkingBlocksBytes(out)
+		if err != nil {
+			return nil, fmt.Errorf("strip unsigned thinking blocks: %w", err)
 		}
 	}
 
@@ -1003,16 +1013,11 @@ func resolveAnthropicOverrides(body []byte, opts EmitOptions) EmitOverrides {
 		ov.StripThinkingBlocks = true
 	}
 
-	if opts.ModelSwitched {
-		ov.StripThinkingBlocks = true
-	}
-
-	// ADD THIS:
 	// If history contains thinking blocks without Anthropic-issued signatures
-	// (synthesised from OSS reasoning_content), strip them — unsigned blocks
-	// cause Anthropic to reject the request with a 400.
+	// (synthesised from OSS reasoning_content), note it so applyOverrides can
+	// strip only the unsigned blocks — leaving valid signed blocks intact.
 	if !ov.StripThinkingBlocks && historyHasUnsignedThinkingBlock(body) {
-		ov.StripThinkingBlocks = true
+		ov.StripUnsignedThinkingBlocks = true
 	}
 
 	if !gjson.GetBytes(body, "max_tokens").Exists() {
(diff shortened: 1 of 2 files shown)


internal/translate/envelope.go
```diff
@@ -1011,6 +1028,98 @@ func resolveAnthropicOverrides(body []byte, opts EmitOptions) EmitOverrides {
...
+			// including signed thinking blocks.
+			if block.Get("type").String() == "thinking" &&
+				block.Get("signature").String() == "" {
+				return true
+			}
+			kept = append(kept, block.Raw)
+			return true
+		})
+
+		filteredContent := "[" + strings.Join(kept, ",") + "]"
```
steventohme, line 1110: This is the one I'd hold on. If a message's only block is an unsigned thinking block, `kept` is empty and this produces `content: []`, which Anthropic also rejects with a 400. So in that case we'd be replacing one 400 with another (it can happen when an OSS turn returns only reasoning and nothing else). Could you drop the whole message when `kept` is empty? The existing `stripThinkingBlocksBytes` has the same problem, so ideally fix both and add a test for the empty case.

internal/translate/envelope.go
```diff
@@ -1011,6 +1028,98 @@ func resolveAnthropicOverrides(body []byte, opts EmitOptions) EmitOverrides {
...
+// conversation history contains a thinking block without a signature field.
+// This occurs when OSS models (DeepSeek, Qwen) return reasoning_content which
+// the router translates into thinking blocks — but without Anthropic-issued
+// signatures. Sending these to Anthropic causes a 400 error.
+func historyHasUnsignedThinkingBlock(body []byte) bool {
+	found := false
+	gjson.GetBytes(body, "messages").ForEach(func(_, msg gjson.Result) bool {
+		msg.Get("content").ForEach(func(_, block gjson.Result) bool {
+			if block.Get("type").String() == "thinking" &&
+				block.Get("signature").String() == "" {
```
steventohme: Minor: checking `signature == ""` means a signature that's only whitespace counts as signed. Unlikely in practice since OSS models leave the field out entirely, but `strings.TrimSpace(...) == ""` would cover it.

internal/translate/envelope.go
```diff
@@ -1011,6 +1028,98 @@ func resolveAnthropicOverrides(body []byte, opts EmitOptions) EmitOverrides {
 	return ov
 }
 
+// historyHasUnsignedThinkingBlock returns true if any message in the
+// conversation history contains a thinking block without a signature field.
+// This occurs when OSS models (DeepSeek, Qwen) return reasoning_content which
+// the router translates into thinking blocks — but without Anthropic-issued
+// signatures. Sending these to Anthropic causes a 400 error.
+func historyHasUnsignedThinkingBlock(body []byte) bool {
```
steventohme, line 1038: Worth a one-line comment that this only matches `type == "thinking"` on purpose and leaves `redacted_thinking` alone, since those don't have a signature by design (they use `data` instead). Helps the next person not "fix" it by mistake.

steventohme (changes requested):
Core logic is right. One change inline before merge (the empty-content case can turn one 400 into another), the rest is small. Also the Bugbot summary up top is stale, the current version already covers its points.

internal/translate/envelope.go
rohith500: Fixed. 

Added `if len(kept) == 0 { return true }` in `stripUnsignedThinkingBlocksBytes` so a message whose only content is unsigned thinking blocks is dropped entirely rather than emitting `content:[]`. 

Added `TestStripUnsignedThinkingBlocks_EmptyContentDropsMessage` covering this case. 

Left `stripThinkingBlocksBytes` unchanged — `TestAnthropicSameFormat_AllThinkingBlocksRemoved` explicitly asserts `content:[]` as the expected contract for that path.

internal/translate/envelope.go
rohith500: Fixed

`strings.TrimSpace(block.Get("signature").String()) == ""` in both the detection loop in`historyHasUnsignedThinkingBlock` and the filter loop in `stripUnsignedThinkingBlocksBytes`.

internal/translate/envelope.go
rohith500: Added: `// Note: only type=="thinking" is matched intentionally; 
redacted_thinking blocks use a "data" field instead of "signature" and must not be stripped.`

rohith500:
Addressed all three review comments:

- **Empty content drops message**: Added `if len(kept) == 0 { return true }` in `stripUnsignedThinkingBlocksBytes` so a message whose only block is an unsigned thinking block is dropped entirely rather than producing `content:[]`. Left `stripThinkingBlocksBytes` unchanged — its existing test explicitly asserts `content:[]` as the contract for that path. Added `TestStripUnsignedThinkingBlocks_EmptyContentDropsMessage` covering the drop case.
- **TrimSpace on signature check**: Applied `strings.TrimSpace(...)` in both `historyHasUnsignedThinkingBlock` and the filter loop in `stripUnsignedThinkingBlocksBytes`.
- **redacted_thinking comment**: Added a note explaining `redacted_thinking` is intentionally excluded since it uses `data` not `signature`.

All 30 packages pass `go test ./... -race -count=1` and `make check`.

PR #450, opened 2026-06-16
fix(providers): buffer Google error body before writing to client

rohith500:
NativeClient.Proxy called CopyUpstreamHeaders + WriteHeader unconditionally before the >= 400 check. This committed the preludeBuffer, making dispatchWithFallback's retry gate (preludeBuf.Committed()) permanently true — so no retry could ever fire on a Google 4xx/5xx error.

Concretely: a transient Google 500 on the Anthropic surface (/v1/messages) streamed raw Google JSON to the client and skipped all configured fallbacks.

Fix: mirror the pattern already used by the Anthropic and OpenAI-compat clients — buffer the error body with readCapped, capture headers into a private http.Header, and return *providers.UpstreamErrorResponse without touching w. The success path writes headers and status as before.

Also fixes the Passthrough method where the same premature WriteHeader prevented clean branch separation.

Tests: 5 subtests covering 429/503/500/400/401 error buffering, header capture, and success path preservation.

### DIFF
--- a/internal/providers/google/native_client.go
+++ b/internal/providers/google/native_client.go
@@ -117,18 +117,16 @@ func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep
 	defer resp.Body.Close()
 	t.StampUpstreamHeaders()
 
-	providers.CopyUpstreamHeaders(w, resp)
-	w.WriteHeader(resp.StatusCode)
-
 	if resp.StatusCode >= 400 {
-		var snip [1024]byte
-		n, _ := io.ReadFull(resp.Body, snip[:])
-		if n > 0 {
+		// Buffer the upstream error — do NOT touch w. The proxy's failover loop
+		// decides whether to retry on another binding or flush this buffer to
+		// the client. Calling w.WriteHeader here would commit the preludeBuffer
+		// and make retries impossible (preludeBuf.Committed() gates all retries).
+		bufBody, totalRead, drainErr := readCapped(resp.Body, providers.MaxBufferedErrorBytes)
+		if len(bufBody) > 0 {
 			t.StampUpstreamFirstByte()
 		}
-		_, snipWriteErr := w.Write(snip[:n])
-		rest, copyErr := io.Copy(w, resp.Body)
-		if copyErr == nil {
+		if drainErr == nil {
 			t.StampUpstreamEOF()
 		}
 		logUpstreamStatus(
@@ -136,16 +134,16 @@ func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep
 			resp.StatusCode,
 			"routed_model", decision.Model,
 			"streaming", stream,
-			"body_preview", string(snip[:n]),
-			"body_total_bytes", int64(n)+rest,
+			"body_preview", previewBytes(bufBody),
+			"body_total_bytes", totalRead,
 		)
-		if snipWriteErr != nil {
-			return snipWriteErr
+		errHeaders := http.Header{}
+		providers.CopyUpstreamHeaders(headerCapture{errHeaders}, resp)
+		return &providers.UpstreamErrorResponse{
+			Status:  resp.StatusCode,
+			Headers: errHeaders,
+			Body:    bufBody,
 		}
-		if copyErr != nil {
-			return copyErr
-		}
-		return &providers.UpstreamStatusError{Status: resp.StatusCode}
 	}
 
 	// Output-progress watchdog: StreamBody's byte-idle watchdog resets on ANY
@@ -204,9 +202,9 @@ func (c *NativeClient) Passthrough(ctx context.Context, prep providers.PreparedR
 	}
 	defer resp.Body.Close()
 
-	providers.CopyUpstreamHeaders(w, resp)
-	w.WriteHeader(resp.StatusCode)
 	if resp.StatusCode >= 400 {
+		providers.CopyUpstreamHeaders(w, resp)
+		w.WriteHeader(resp.StatusCode)
 		var snip [1024]byte
 		n, _ := io.ReadFull(resp.Body, snip[:])
 		_, snipWriteErr := w.Write(snip[:n])
@@ -226,6 +224,8 @@ func (c *NativeClient) Passthrough(ctx context.Context, prep providers.PreparedR
 		}
 		return &providers.UpstreamStatusError{Status: resp.StatusCode}
 	}
+	providers.CopyUpstreamHeaders(w, resp)
+	w.WriteHeader(resp.StatusCode)
 	_, err = io.Copy(w, resp.Body)
 	return err
 }
@@ -250,5 +250,39 @@ func (c *NativeClient) applyAPIKey(ctx context.Context, req *http.Request) {
 		req.Header.Set("x-goog-api-key", c.apiKey)
 	}
 }
+
+// readCapped reads up to limit bytes from r, then drains up to 1 MiB without
+// retention so the upstream connection is released promptly. Returns the
+// buffered prefix, total bytes read across both reads, and any error (io.EOF
+// mapped to nil).
+func readCapped(r io.Reader, limit int) ([]byte, int64, error) {
+	prefix, err := io.ReadAll(io.LimitReader(r, int64(limit)))
+	totalRead := int64(len(prefix))
+	if err != nil {
+		return prefix, totalRead, err
+	}
+	const maxDrain = 1 << 20 // 1 MiB
+	rest, drainErr := io.Copy(io.Discard, io.LimitReader(r, maxDrain))
+	totalRead += rest
+	return prefix, totalRead, drainErr
+}
+
+// previewBytes returns up to the first 1 KiB of body as a string for logging.
+func previewBytes(body []byte) string {
+	const previewLimit = 1024
+	if len(body) > previewLimit {
+		return string(body[:previewLimit])
+	}
+	return string(body)
+}
+
+// headerCapture is a minimal http.ResponseWriter that captures headers only.
+// Used to reuse providers.CopyUpstreamHeaders against an http.Header we own.
+// Write and WriteHeader are no-ops.
+type headerCapture struct{ h http.Header }
+
+func (c headerCapture) Header() http.Header       { return c.h }
+func (c headerCapture) Write([]byte) (int, error) { return 0, nil }
+func (c headerCapture) WriteHeader(int)           {}
 
 var _ providers.Client = (*NativeClient)(nil)
--- a/internal/providers/google/native_client.go
+++ b/internal/providers/google/native_client.go
@@ -163,6 +163,8 @@ func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep
 		}
 	}
 
+	providers.CopyUpstreamHeaders(w, resp)
+	w.WriteHeader(resp.StatusCode)
 	return httputil.StreamBody(ctx, cancel, c.idleTimeout(), resp.Body, resp.StatusCode, w, t)
 }
 
--- a/internal/providers/google/native_client.go
+++ b/internal/providers/google/native_client.go
@@ -122,7 +122,7 @@ func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep
 		// decides whether to retry on another binding or flush this buffer to
 		// the client. Calling w.WriteHeader here would commit the preludeBuffer
 		// and make retries impossible (preludeBuf.Committed() gates all retries).
-		bufBody, totalRead, drainErr := readCapped(resp.Body, providers.MaxBufferedErrorBytes)
+		bufBody, totalRead, drainErr := providers.ReadCapped(resp.Body, providers.MaxBufferedErrorBytes)
 		if len(bufBody) > 0 {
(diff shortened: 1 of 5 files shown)


internal/providers/google/native_client.go
```diff
@@ -194,4 +196,38 @@ func (c *NativeClient) applyAPIKey(ctx context.Context, req *http.Request) {
 	}
 }
 
+// readCapped reads up to limit bytes from r, then drains up to 1 MiB without
+// retention so the upstream connection is released promptly. Returns the
+// buffered prefix, total bytes read across both reads, and any error (io.EOF
+// mapped to nil).
+func readCapped(r io.Reader, limit int) ([]byte, int64, error) {
```
steventohme: `readCapped`, `previewBytes`, and `headerCapture` are copy-pasted verbatim from `openaicompat/client.go:166-200`. None of them has Google-specific logic; they only depend on `providers.MaxBufferedErrorBytes` and `providers.CopyUpstreamHeaders`, which are already shared.

This is the second client to need this, so it should move into the `providers` package instead of being duplicated. `headerCapture` can go away entirely if we add a `providers.CaptureUpstreamHeaders(resp) http.Header` helper. The two copies of `readCapped` have already drifted (this doc comment says "released promptly", the `openaicompat` one says "to bound failover latency on a slow upstream"), which is the concrete argument for keeping one copy.

internal/providers/google/native_client.go
```diff
@@ -147,9 +147,9 @@ func (c *NativeClient) Passthrough(ctx context.Context, prep providers.PreparedR
 	}
 	defer resp.Body.Close()
 
-	providers.CopyUpstreamHeaders(w, resp)
-	w.WriteHeader(resp.StatusCode)
 	if resp.StatusCode >= 400 {
+		providers.CopyUpstreamHeaders(w, resp)
+		w.WriteHeader(resp.StatusCode)
```
steventohme, line 207: `Passthrough` still flushes to `w` and returns `UpstreamStatusError`, so it doesn't participate in buffered failover and there's no prelude buffer to commit here. Moving `CopyUpstreamHeaders` + `WriteHeader` into each branch is behaviorally identical to before. This is fine as cleanup but it isn't part of the fix, so it shouldn't be described as one in the commit message.

internal/providers/google/native_client_test.go
```diff
@@ -122,3 +122,179 @@ func TestNativeClient_DefaultBaseURL(t *testing.T) {
...
+			prep := providers.PreparedRequest{Body: []byte(`{"contents":[]}`), Headers: make(http.Header)}
+
+			err := c.Proxy(context.Background(), router.Decision{Model: "gemini-x"}, prep, rec,
+				httptest.NewRequest(http.MethodPost, "/v1/messages", strings.NewReader("")))
+
+			// Must return *UpstreamErrorResponse (buffered), never *UpstreamStatusError (flushed).
+			var buffered *providers.UpstreamErrorResponse
+			require.ErrorAs(t, err, &buffered, "Proxy must return UpstreamErrorResponse on %d", tc.upstreamStatus)
+			assert.Equal(t, tc.upstreamStatus, buffered.Status)
+			assert.Contains(t, string(buffered.Body), tc.upstreamBody[:20])
```
steventohme: `tc.upstreamBody[:20]` panics if a future case has a body shorter than 20 bytes, and the offset is arbitrary. The body is small and fully buffered, so `assert.Equal(t, tc.upstreamBody, string(buffered.Body))` is simpler and stricter.

internal/providers/google/native_client_test.go
```diff
@@ -122,3 +122,179 @@ func TestNativeClient_DefaultBaseURL(t *testing.T) {
...
+	require.ErrorAs(t, err, &buffered)
+	assert.Equal(t, "30", buffered.Headers.Get("Retry-After"),
+		"upstream Retry-After must be captured in UpstreamErrorResponse.Headers")
+	assert.Empty(t, rec.Header().Get("Retry-After"),
+		"upstream error headers must not reach the client ResponseWriter")
+}
+
+// TestNativeClient_Proxy_2xxWritesDirectly verifies the success path is
+// unchanged: 2xx responses are still written to w normally.
+func TestNativeClient_Proxy_2xxWritesDirectly(t *testing.T) {
```
steventohme, line 226: The success-path test uses a non-streaming 200. The branch that actually matters for Gemini is `:streamGenerateContent?alt=sse` going through `StreamBody`, which needs `stream=true` via the `GeminiStreamHintHeader`. There's also no test for `readCapped` truncating a body over `MaxBufferedErrorBytes`. The new helper ships untested.

internal/providers/google/native_client_test.go
```diff
@@ -122,3 +122,179 @@ func TestNativeClient_DefaultBaseURL(t *testing.T) {
...
+			assert.Equal(t, tc.upstreamStatus, buffered.Status)
+			assert.Contains(t, string(buffered.Body), tc.upstreamBody[:20])
+
+			// The client ResponseWriter must be completely untouched.
+			assert.Equal(t, http.StatusOK, rec.Code,
+				"WriteHeader must not be called on the error path (preludeBuffer commit prevention)")
+			assert.Empty(t, rec.Body.String(),
+				"no bytes must be written to the client on the error path")
+
+			assert.Equal(t, tc.wantRetryable, providers.IsRetryable(err),
```
steventohme: `wantRetryable` is asserting on `providers.IsRetryable`, which lives in another package and is tested there. It doesn't exercise anything the Google client controls, so it adds noise to a test whose stated purpose is "ErrorBufferedNotFlushed."

steventohme (changes requested):
The fix is correct: deferring the writes and returning *UpstreamErrorResponse is the right way to keep the failover gate open on Google errors. Two things in the description and a few smaller items inline.

The description says this mirrors "the pattern already used by the Anthropic and OpenAI-compat clients." Only `openaicompat` buffers. The Anthropic client (`anthropic/client.go:82`) does the unconditional `CopyUpstreamHeaders` + `WriteHeader` before the `>= 400` check, same as the code this PR is fixing. Worth correcting the description.

Follow-up, not this PR: `anthropic/client.go` has the same failover bug being fixed here. It commits the response before the status check, so `dispatchWithFallback` can't retry on an Anthropic upstream error either. Should file an issue.

rohith500:
Thanks for the thorough review — pushed 137376a addressing the test feedback, and replying inline on the rest.

**PR description / "mirrors Anthropic and OpenAI-compat"** — you're right that this overstated it. `anthropic/client.go`'s `Proxy` does check `>= 400` before touching `w` (so the comparison holds for `Proxy`), but I conflated that with `Passthrough`, which I hadn't actually checked closely. Will edit the description to be specific: this mirrors `openaicompat/client.go`'s pattern, and `anthropic/client.go`'s `Proxy` independently does the same thing correctly.

**Anthropic's Passthrough bug** — confirmed, same ordering issue (`CopyUpstreamHeaders` + `WriteHeader` before the `>= 400` check). Agreed this should be its own issue rather than folded in here since it's an unrelated client. Will file it.

**readCapped/previewBytes/headerCapture duplication** — agreed on the substance, and good catch on the comment drift (the "released promptly" vs. "bound failover latency" wording really did drift apart, which is exactly the failure mode of duplicating instead of sharing). I'd rather not widen this PR's diff into `openaicompat/client.go` while it's already mid-review, so I'm deferring this to a focused follow-up that moves all three into `providers` and updates both call sites in one pass. Open to doing it in this PR instead if you'd prefer that over a follow-up.

**Passthrough reordering described as part of the fix** — fair, it's behaviorally identical before/after for Passthrough specifically (always writes to w, always returns `*UpstreamStatusError`). I'll tighten the commit message to describe it as cleanup for consistency with Proxy's branching, not as part of the functional fix.

**Test fixes** — done in 137376a:
- `tc.upstreamBody[:20]` → full-body `assert.Equal` (the slice panicked for any body under 20 bytes; full comparison is simpler and stricter given the body is small and fully buffered anyway).
- Dropped the `providers.IsRetryable` assertion — agreed it's noise in a test about buffering, not retryability, and belongs to another package's test suite.
- Added `Proxy_StreamingSuccessWritesViaStreamBody` covering the `:streamGenerateContent?alt=sse` / `StreamBody` path, which had no coverage before.
- Added `Proxy_ErrorBodyTruncatedAtCap` covering `readCapped` actually truncating at `MaxBufferedErrorBytes` rather than buffering an oversized body in full.

Let me know if you'd rather have the `providers` package consolidation in this PR instead of a follow-up — happy to do either.

internal/providers/google/native_client.go
rohith500: it's identical before and after for Passthrough since it always writes to w either way. Will describe this as cleanup rather than part of the fix going forward.

internal/providers/google/native_client.go
rohith500: The comment drift you caught is good evidence for why. I'm deferring the actual move into the providers package to a focused follow-up rather than widening this PR's diff into openaicompat/client.go while it's already mid-review. Happy to do it here instead if you'd rather.

rohith500:
@steventohme 

Went ahead and did the `providers` package consolidation in this PR after all, in four separate commits so each is easy to review on its own:

- `37942ee` — adds `providers.ReadCapped`, `providers.PreviewBytes`, and `providers.CaptureUpstreamHeaders(resp) http.Header` to `internal/providers/provider.go`. This last one replaces the `headerCapture`-plus-empty-`http.Header` pattern every call site repeated; `headerCapture` is now a private implementation detail of `CaptureUpstreamHeaders` rather than something each client package redefines.
- `867667a` — switches `google/native_client.go` over, removing its local copies.
- `1748b22` — switches `openaicompat/client.go` over, removing its local copies.
- `c661b08` — switches `anthropic/client.go` over, removing its local copies.

All three client packages' existing tests pass unmodified, and `make check` is green across the whole module. No behavior change anywhere — this is purely the deduplication you flagged.

Let me know if you'd rather see this split into its own PR instead of bundled here; happy to cherry-pick these four commits onto a fresh branch off `main` if that's cleaner for review.

internal/providers/google/native_client.go
```diff
@@ -79,37 +79,35 @@ func (c *NativeClient) Proxy(ctx context.Context, decision router.Decision, prep
...
 			"routed_model", decision.Model,
 			"streaming", stream,
-			"body_preview", string(snip[:n]),
-			"body_total_bytes", int64(n)+rest,
+			"body_preview", providers.PreviewBytes(bufBody),
+			"body_total_bytes", totalRead,
 		)
-		if snipWriteErr != nil {
-			return snipWriteErr
+		return &providers.UpstreamErrorResponse{
```
steventohme, line 140: One thing for a follow-up: when all bindings are exhausted, this buffered body gets translated by `OpenAIToAnthropicError`, but it's now a raw Gemini error, which puts the error type in `error.status` instead of `error.type`. So the client gets the right message but an empty type. A `GeminiToAnthropicError` helper would fix that. Fine to do separately, just flagging it.

internal/providers/provider.go
```diff
@@ -141,6 +142,53 @@ func (e *UpstreamErrorResponse) Error() string {
 // of the upstream stream is drained without retention.
 const MaxBufferedErrorBytes = 64 * 1024
 
+// ReadCapped reads up to limit bytes from r into a buffer, then drains the
+// rest without retention up to a 1 MiB cap so the upstream connection is
+// released promptly and failover latency on a slow upstream stays bounded.
+// Returns the buffered prefix, total bytes read across both reads, and any
+// error (io.EOF mapped to nil).
+func ReadCapped(r io.Reader, limit int) ([]byte, int64, error) {
```
steventohme, line 179: No problem with moving `ReadCapped`/`PreviewBytes`/`CaptureUpstreamHeaders` here, it's clean and removes the duplicated copies. Only note is that it's most of what makes this PR large. Happy to keep it, or you could split it into its own PR to keep this one focused. Up to you.

steventohme (dismissed):

internal/providers/google/native_client.go
rohith500: Good catch on the error.status vs error.type mismatch. Gemini uses error.status ("RESOURCE_EXHAUSTED", "INVALID_ARGUMENT", etc.) where OpenAI uses error.type, so OpenAIToAnthropicError produces the right message but an empty type field on any exhausted-bindings Gemini error. 

This was only reachable after the buffering fix in this PR — before it, Gemini errors streamed raw to the client and never went through translation. Filed #487 to track a GeminiToAnthropicError helper that maps Gemini status strings to Anthropic error types.

internal/providers/provider.go
rohith500: Happy to keep the ReadCapped / PreviewBytes / CaptureUpstreamHeaders consolidation in this PR. 

The motivation for doing it here rather than a follow-up is exactly what you flagged in the review — the two existing copies had already drifted in their doc comments ("released promptly" vs "bound failover latency"), which is the concrete failure mode of duplicating instead of sharing. 

Now that a third client needed the same helpers, moving them into providers in the same PR as the 
fix felt cleaner than leaving three copies and filing a deferred cleanup. All three client packages' existing tests pass unmodified and make check is green across the whole module.

PR #461, opened 2026-06-18
fix(admin/ui): reject env-keyed provider writes at backend; don't fail open in ProviderKeysPanel

### LINKED ISSUE #459: Provider key write endpoint has no env-key check — admin dashboard config failure allows a BYOK key to silently shadow the deployment's env-configured Anthropic/OpenRouter/Google credentials
This deployment can authenticate to upstream providers (Anthropic, OpenRouter, Google, etc.) in two ways: an environment variable set at deploy time (e.g. `ANTHROPIC_API_KEY`), or a key entered through the admin dashboard and stored in the database ("BYOK" — bring your own key). The dashboard is supposed to prevent BYOK writes for any provider that already has an env key, by graying out that provider's field and showing an "ENV" badge.

That protection lives entirely in the frontend, is derived from a single network call, and has no backend equivalent. If that one call fails — for any ordinary, content-unrelated reason — the dashboard silently treats every env-keyed provider as if it had never been configured, and the backend write endpoint accepts a new key for it with no validation at all. Because of how credential resolution is implemented elsewhere in the system, the new (possibly accidental, possibly wrong) key then becomes the credential actually used for that provider going forward — silently, with no error, no diff, and no UI indication that a swap occurred.

This was reproduced live end-to-end on June 18, 2026, including watching a deliberately fake canary key get presented to Anthropic's real production API and rejected by Anthropic's own auth layer — direct proof the wrong credential was actually used for a live outbound call, not just sitting unused in the database.

### Piece 1 — internal/api/admin/config.go, ConfigHandler (correct, but isolated)

This handler builds the response for `GET /admin/v1/config`. For each known provider `p`, it checks whether the provider's env var is set, using this logic:

if config.GetOr(providers.APIKeyEnvVar(p), "") != "" { envKeyed = append(envKeyed, p) }

`providers.APIKeyEnvVar(p)` maps a provider name to its expected env var name (e.g. "anthropic" maps to "ANTHROPIC_API_KEY"). If that env var is set, the provider's name goes into the `env_provider_keys` array sent back to the frontend. This is the only place in the entire backend that performs this check. It exists purely to tell the UI what to gray out, and is never reused anywhere else.

### Piece 2 — internal/api/admin/keys.go, UpsertExternalKeyHandler (the actual bug — missing guard)

This is the handler behind `POST /admin/v1/provider-keys` (the "Save key" button). It takes the provider and key from the request body and calls into `authSvc.UpsertExternalAPIKey(...)` to persist them. There is no call to `config.GetOr(...)`, no reference to `providers.APIKeyEnvVar`, and no check of any kind against env state anywhere in this handler. It will accept and persist a key for "anthropic" even when `ANTHROPIC_API_KEY` is already set in the environment. The exact one-line check that exists in Piece 1 was simply never added here. This is the root cause: the validation logic exists in the codebase, but only on the read path, never on the write path.

### Piece 3 — frontend settings/page.tsx, ProviderKeysPanel (what makes the bug trivially reachable)

This component fetches `/admin/v1/config` on mount and reads the `env_provider_keys` array to decide which providers get the "ENV" badge and a disabled input. Its failure handling on that fetch does not surface an error state or retry — when the request fails, the component's catch path leaves the env-keyed list empty, so every provider renders as fully editable, visually indistinguishable from a provider that has genuinely never been configured. This turns "the write endpoint has no check" from a theoretical API-level gap into something reachable by clicking a normal UI form, triggered by something as mundane as a flaky network connection.

### Why the wrong key then wins — proxy/credentials.go, resolveAndInjectCredentials (around lines 2117-2130)

When the router dispatches a request to a provider, this function decides which credential to use. It checks BYOK (database-stored, per-installation) keys first, and only falls back to the environment key if no BYOK key exists for that provider. This ordering is deliberate and necessary: in the system's managed-mode deployment, there is no environment key at all, and BYOK is the only mechanism by which any customer's credentials ever get supplied. This priority order is correct in general and is not itself a bug. The bug is that nothing stops a BYOK row from being created by accident, for a provider where the same deployment's own env var is meant to be the sole source of truth.

## Live, step-by-step reproduction

Step 1, baseline. The Settings page showed Anthropic, OpenRouter, and Google each with an ENV badge and a disabled key field. Confirmed via the Network tab that this came from `GET /admin/v1/config` returning `env_provider_keys: ["anthropic", "openrouter", "google"]`.

Step 2, simulate the transient failure. Used Chrome DevTools "Block request URL" on `http://localhost:3000/admin/v1/config`. This simulates an ordinary network failure — no exploit, no special request, just making one endpoint temporarily unreachable.

Step 3, reload and observe the failure mode. All config requests on the page (4 per load, since ProviderKeysPanel and a separate ConfigPanel each fetch it independently, doubled again by React dev-mode effect invocation) showed as blocked in the Network tab. The "Active provider keys" list changed from showing three ENV-badged entries to "No provider keys configured." The "Add a key" provider dropdown, previously limited to OpenAI, now listed Anthropic as a normal, selectable option.

Step 4, trigger the write with no backend validation. Entered an intentionally fake value, a fabricated canary string never used as a real credential, into the now-unlocked Anthropic field and clicked "Save key." The request succeeded with no error response. The dashboard then displayed two separate "Anthropic" rows: the original ENV-badged entry, and a second, unbadged entry holding the canary key, with no warning, merge, or conflict indicator between them.

Step 5, confirm it's not just a UI artifact. Removed the DevTools block on `/admin/v1/config` and reloaded. All four config requests now returned 200. The duplicate Anthropic row persisted unchanged — the system does not detect or reconcile having two stored credentials for the same provider once both exist.

Step 6, force a live dispatch to Anthropic to observe which key gets used. Sent a request to `/v1/messages` with model claude-opus-4-8. The router's own scoring logic initially chose other providers (Google, then OpenRouter) ahead of Anthropic for this request shape. Temporarily used `PUT /admin/v1/excluded-models` to exclude those specific higher-scoring models, purely so the dispatch decision would land on Anthropic and the outbound call could be directly observed, then reverted this immediately afterward (confirmed via a follow-up check showing an empty exclusion list again). This step is unrelated to the bug itself; it only exists to make the live proof observable directly rather than waiting for organic traffic to land on Anthropic.

Step 7, the proof. With `X-Router-Provider: anthropic` and `X-Router-Model: claude-opus-4-8` confirmed in the response headers, the actual response received was an HTTP 401 Unauthorized, with a Set-Cookie header scoped to domain api.anthropic.com, a Server header of cloudflare, a Strict-Transport-Security header, and a JSON body reading: type error, error type authentication_error, message invalid x-api-key.

The api.anthropic.com cookie domain, the Cloudflare server header, and the HSTS header together confirm this is a genuine response from Anthropic's real production API, not a local stub or proxy-level rejection. Anthropic's own authentication layer received and rejected the canary key. Since the canary key is the only Anthropic credential saved to this installation during the test, this is direct, first-hand proof that it, not the real env key, was the credential actually transmitted on a live outbound request.

## Impact

A single, content-unrelated failure of `GET /admin/v1/config`, something as mundane as a slow network response or a momentary backend hiccup, is sufficient to disable the only existing safeguard against overwriting a self-hosted deployment's env-configured provider credentials. Once a key is saved this way, whether by an admin's accident, a misclick, or a stray automation, it silently becomes the credential used for that provider on every subsequent request, with no error surfaced anywhere and no way to detect the swap from the dashboard short of noticing a duplicate row in the provider key list.

## Proposed fix

Backend fix, in internal/api/admin/keys.go, UpsertExternalKeyHandler — this is the fix that actually closes the hole. Add the same check already present in ConfigHandler, before the call to UpsertExternalAPIKey: check whether `config.GetOr(providers.APIKeyEnvVar(req.Provider), "")` is non-empty, and if so, abort the request with a 409 Conflict status and a message such as "Provider already configured via deployment environment variable. Remove the env var before adding a dashboard key." This is a pure environment-variable read with no network call and no dependency on `/admin/v1/config` having succeeded, so it closes the gap completely regardless of frontend state.

Frontend fix, in settings/page.tsx, ProviderKeysPanel — defense in depth, not sufficient by itself. The component should not fail open when `GET /admin/v1/config` errors. It should surface an explicit error state and keep the "Add a key" form disabled until the config fetch succeeds, rather than defaulting to treating every provider as unconfigured. This alone would not close the hole for direct API access such as curl or scripts, which is why the backend fix above is the one that matters; this only prevents the failure mode from being reachable through ordinary dashboard use.

rohith500:
## What

Fixes #459 — two cooperating fixes for the same vulnerability:

**Backend (internal/api/admin/keys.go):** `UpsertExternalKeyHandler` had no guard against writing a BYOK key for a provider already configured via a deployment env var (e.g. `ANTHROPIC_API_KEY`). The only protection was UI-level. Add `isEnvKeyed(provider)` — a pure env-var read mirroring the check `ConfigHandler` already uses to populate `EnvProviderKeys` — and call it before `UpsertExternalAPIKey`. Returns 409 Conflict when the provider is env-keyed, regardless of whether the config endpoint is healthy.

**Frontend (frontend/src/app/(app)/settings/page.tsx):** `ProviderKeysPanel` initialised `envKeyed` as `[]` and silently fell back to `[]` on a failed `GET /admin/v1/config`, making every provider appear available for a BYOK write when config was transiently unreachable. `envKeyed` now starts as `null` (not yet loaded), a separate `configError` state replaces the silent catch, and the Add-a-key card is not rendered at all until `configReady`.

## Why this approach

The backend fix is the authoritative guard — it closes the hole regardless of frontend state or direct API access. The frontend fix removes the UI path that made the gap trivially reachable through normal dashboard use. This matches the project's existing stance against fail-open behavior (CONTRIBUTING.md: "Silent fallbacks... PRs re-introducing fail-open behavior will be rejected").

## How it was tested

- `make check` green (generate + vet + build + full test suite)
- Three new unit tests in `internal/api/admin/keys_internal_test.go` covering: env var set → rejected, env var unset → allowed, unknown provider → allowed
- Live end-to-end: blocked `GET /admin/v1/config` in DevTools — confirmed Add-a-key card no longer renders while config is unknown, error banner shown instead
- Live end-to-end: `curl POST /admin/v1/provider-keys` with `ANTHROPIC_API_KEY` set → confirmed 409 Conflict with correct error message
- Live end-to-end: unblocked config → confirmed ENV badges restore correctly, no regression in working case

### DIFF
--- a/frontend/src/app/(app)/settings/page.tsx
+++ b/frontend/src/app/(app)/settings/page.tsx
@@ -353,7 +353,9 @@ function providerLabel(p: Provider): string {
 
 function ProviderKeysPanel() {
   const [keys, setKeys] = useState<ExternalKey[]>([]);
-  const [envKeyed, setEnvKeyed] = useState<Provider[]>([]);
+  // null = config not yet loaded; [] = loaded with no env-keyed providers
+  const [envKeyed, setEnvKeyed] = useState<Provider[] | null>(null);
+  const [configError, setConfigError] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
   const [pickedProvider, setPickedProvider] = useState<Provider | null>(null);
   const [keyValue, setKeyValue] = useState("");
@@ -380,16 +382,18 @@ function ProviderKeysPanel() {
           (PROVIDERS as readonly string[]).includes(p),
         );
         setEnvKeyed(set);
+        setConfigError(null);
       })
       .catch(() => {
-        // Non-fatal: the panel still works, env-keyed providers just won't
-        // be flagged as read-only.
-        setEnvKeyed([]);
+        setConfigError("Couldn't verify provider configuration — refresh to try again");
       });
   }, []);
 
-  const taken = new Set<string>([...keys.map(k => k.provider), ...envKeyed]);
-  const available: Provider[] = PROVIDERS.filter(p => !taken.has(p));
+  const configReady = envKeyed !== null && configError === null;
+  const taken = new Set<string>([...keys.map(k => k.provider), ...(envKeyed ?? [])]);
+  // Only offer providers for addition once we know which are env-keyed; until
+  // then any of the four could be read-only and we must not silently offer them.
+  const available: Provider[] = configReady ? PROVIDERS.filter(p => !taken.has(p)) : [];
   const provider: Provider | null =
     pickedProvider != null && available.includes(pickedProvider)
       ? pickedProvider
@@ -423,10 +427,11 @@ function ProviderKeysPanel() {
     }
   }
 
-  const hasAnyKey = keys.length > 0 || envKeyed.length > 0;
+  const hasAnyKey = keys.length > 0 || (envKeyed ?? []).length > 0;
   return (
     <>
       {error && <ErrorBanner>{error}</ErrorBanner>}
+      {configError && <ErrorBanner>{configError}</ErrorBanner>}
 
       {available.length > 0 && provider != null ? (
         <Card>
@@ -485,7 +490,7 @@ function ProviderKeysPanel() {
           </Card.Header>
           <Card.Content>
             <ul className="divide-y divide-border">
-              {envKeyed.map(p => (
+              {(envKeyed ?? []).map(p => (
                 <li
                   key={`env-${p}`}
                   className="flex items-center justify-between px-5 py-3"
--- a/internal/api/admin/keys.go
+++ b/internal/api/admin/keys.go
@@ -3,6 +3,7 @@ package admin
 import (
 	"errors"
 	"net/http"
+	"strings"
 	"time"
 
 	"workweave/router/internal/auth"
@@ -208,7 +209,7 @@ func ListExternalKeysHandler(authSvc *auth.Service) gin.HandlerFunc {
 // write path can enforce it independently of whether /admin/v1/config ever
 // succeeds.
 func isEnvKeyed(provider string) bool {
-	return config.GetOr(providers.APIKeyEnvVar(provider), "") != ""
+	return config.GetOr(providers.APIKeyEnvVar(strings.ToLower(provider)), "") != ""
 }
 
 func UpsertExternalKeyHandler(authSvc *auth.Service) gin.HandlerFunc {
--- a/internal/api/admin/keys.go
+++ b/internal/api/admin/keys.go
@@ -4,8 +4,9 @@ import (
 	"errors"
 	"net/http"
 	"time"
-
 	"workweave/router/internal/auth"
+	"workweave/router/internal/config"
+	"workweave/router/internal/providers"
 
 	"github.com/gin-gonic/gin"
 )
@@ -200,6 +201,15 @@ func ListExternalKeysHandler(authSvc *auth.Service) gin.HandlerFunc {
 	}
 }
 
+// isEnvKeyed reports whether provider's upstream API key is supplied via a
+// deployment env var (e.g. ANTHROPIC_API_KEY). Mirrors the check ConfigHandler
+// uses to populate EnvProviderKeys — kept here as its own function so the
+// write path can enforce it independently of whether /admin/v1/config ever
+// succeeds.
+func isEnvKeyed(provider string) bool {
+	return config.GetOr(providers.APIKeyEnvVar(provider), "") != ""
+}
+
 func UpsertExternalKeyHandler(authSvc *auth.Service) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		installation, ok := resolveInstallation(c, authSvc)
@@ -209,6 +219,10 @@ func UpsertExternalKeyHandler(authSvc *auth.Service) gin.HandlerFunc {
 		var req upsertExternalKeyRequest
 		if err := c.ShouldBindJSON(&req); err != nil {
 			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "Provider and key are required."})
+			return
+		}
+		if isEnvKeyed(req.Provider) {
+			c.AbortWithStatusJSON(http.StatusConflict, gin.H{"error": "Provider already configured via deployment environment variable. Remove the env var before adding a dashboard key."})
 			return
 		}
 		key, err := authSvc.UpsertExternalAPIKey(c.Request.Context(), installation.ID, req.Provider, req.Key, req.Name, nil)
--- a/internal/api/admin/keys.go
+++ b/internal/api/admin/keys.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"net/http"
 	"time"
+
 	"workweave/router/internal/auth"
 	"workweave/router/internal/config"
 	"workweave/router/internal/providers"
--- a/internal/api/admin/keys_internal_test.go
+++ b/internal/api/admin/keys_internal_test.go
@@ -18,6 +18,11 @@ func TestIsEnvKeyed_TrueWhenEnvVarSet(t *testing.T) {
 	assert.True(t, isEnvKeyed("anthropic"))
 }
 
+func TestIsEnvKeyed_TrueWhenProviderIsMixedCase(t *testing.T) {
+	t.Setenv("ANTHROPIC_API_KEY", "sk-ant-test-value")
+	assert.True(t, isEnvKeyed("Anthropic"))
+	assert.True(t, isEnvKeyed("ANTHROPIC"))
+}
 func TestIsEnvKeyed_FalseWhenEnvVarUnset(t *testing.T) {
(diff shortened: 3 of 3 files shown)


internal/api/admin/keys.go
```diff
@@ -211,6 +221,10 @@ func UpsertExternalKeyHandler(authSvc *auth.Service) gin.HandlerFunc {
 			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "Provider and key are required."})
 			return
 		}
+		if isEnvKeyed(req.Provider) {
```
steventohme, line 226: This is the check that actually closes the hole, so it'd be good to have a test on it. The helper is covered but this handler path isn't. Could you add an httptest + `gin.New()` test that confirms this returns 409 and doesn't call `UpsertExternalAPIKey` when the provider is env-keyed? `catalog_test.go` has the pattern to copy.

internal/api/admin/keys.go
```diff
@@ -4,8 +4,9 @@ import (
 	"errors"
 	"net/http"
 	"time"
-
 	"workweave/router/internal/auth"
```
steventohme, line 9: Small nit: the blank line between the standard library imports and the `workweave/router` imports got removed here. gofmt won't flag it, but the rest of the package keeps them separated, so worth putting back.

steventohme (dismissed):

internal/api/admin/keys.go
rohith500: Added TestUpsertExternalKeyHandler_RejectsEnvKeyedProvider in keys_internal_test.go. It injects a fake installation via a gin middleware shim so resolveInstallation passes, then confirms the handler returns 409 and never reaches UpsertExternalAPIKey when the provider is env-keyed.

internal/api/admin/keys.go
rohith500: Restored —> awk edit I used to add the blank line introduced a formatting issue that gofmt caught; fixed and re-verified with make check.

internal/api/admin/keys.go
```diff
@@ -211,6 +222,10 @@ func UpsertExternalKeyHandler(authSvc *auth.Service) gin.HandlerFunc {
 			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "Provider and key are required."})
 			return
 		}
+		if isEnvKeyed(req.Provider) {
+			c.AbortWithStatusJSON(http.StatusConflict, gin.H{"error": "Provider already configured via deployment environment variable. Remove the env var before adding a dashboard key."})
+			return
```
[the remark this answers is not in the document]
rohith500: Fixed in the latest commit "3055bdf"

isEnvKeyed now calls strings.ToLower(provider) before the map lookup, so "Anthropic", "ANTHROPIC", etc. all hit the guard correctly. Added TestIsEnvKeyed_TrueWhenProviderIsMixedCase to cover both cases.

frontend/src/app/(app)/settings/page.tsx
```diff
@@ -380,16 +382,18 @@ function ProviderKeysPanel() {
           (PROVIDERS as readonly string[]).includes(p),
         );
         setEnvKeyed(set);
+        setConfigError(null);
       })
       .catch(() => {
-        // Non-fatal: the panel still works, env-keyed providers just won't
-        // be flagged as read-only.
-        setEnvKeyed([]);
+        setConfigError("Couldn't verify provider configuration — refresh to try again");
```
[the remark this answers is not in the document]
rohith500: Good catch. 

Looking at the useEffect, it doesn't use an AbortController, so a slow earlier fetch that fails after a faster later one succeeds would incorrectly set configError and hide the Add-a-key flow. I'll add abort/cleanup logic to the effect so stale responses are ignored.

rohith500:
Follow-up changes pushed after initial review (commits 5c4e97e, 3055bdf): Addressed both review comments from @steventohme and two Cursor Bugbot findings:

-- httptest for handler path (5c4e97e): Added TestUpsertExternalKeyHandler_RejectsEnvKeyedProvider in keys_internal_test.go. Injects a fake installation via a gin middleware shim ("router_installation" context key) so resolveInstallation passes auth, then confirms the handler returns 409 and never reaches UpsertExternalAPIKey when the provider is env-keyed.

-- Import blank line (5c4e97e): Restored the stdlib / workweave/router import separator that was accidentally dropped; caught by gofmt during make check.

-- Case-sensitivity bypass (3055bdf): isEnvKeyed now calls strings. ToLower(provider) before the map lookup — a POST with "Anthropic" or "ANTHROPIC" previously bypassed the 409 guard. Added TestIsEnvKeyed_TrueWhenProviderIsMixedCase to cover both variants.

-- Stale config fetch : Valid finding — the useEffect lacks an AbortController, so a slow failing fetch can overwrite configError after a successful one. Deferring to a follow-up issue; the race window is narrow and the null initial state already protects the Add-a-key card on first load.

make check green on all 4 commits.

PR #524, opened 2026-06-26
[codex] Add TrustedRouter upstream provider

jperla:
## Summary

Adds TrustedRouter as a first-class OpenAI-compatible upstream provider alongside OpenRouter.

- Adds `providers.ProviderTrustedRouter`, `TRUSTEDROUTER_API_KEY`, and default base URL `https://api.trustedrouter.com/v1`
- Registers TrustedRouter in the router composition root using the existing OpenAI-compatible adapter
- Treats TrustedRouter as an OpenRouter-compatible aggregate for provider/reasoning hints and session-affinity headers
- Adds TrustedRouter fallback catalog bindings for the existing OpenRouter-backed OSS model pool
- Adds dashboard/provider-key UI support and a reversible migration for saved TrustedRouter BYOK keys
- Documents the env vars in `.env.example`, README, and configuration docs

## Why

TrustedRouter exposes an OpenAI-compatible API, so self-hosted WorkWeave Router deployments can use it as an aggregate provider with the same integration shape as OpenRouter while keeping direct-provider routing behavior unchanged when direct provider keys are present.

## Validation

- `go test ./internal/providers ./internal/providers/openaicompat ./internal/proxy ./internal/router/catalog ./internal/translate ./internal/api/admin`
- `go test ./...`
- `cd frontend && npx tsc --noEmit`

`npm run lint` is currently blocked by the repo's existing frontend setup: `next lint` starts the Next.js ESLint configuration wizard because no ESLint config is present in the clone. I did not let the wizard mutate repo config in this PR.

## Notes

`make migrate-create` could not run locally because the `migrate` binary is not installed, so migration `0028_add-trustedrouter-provider` was created manually following the repo's migration format.


jperla:
Thanks for the automated review. The only blocker I see is the allowed-authors policy rather than a code issue. If this needs to be opened by a maintainer or converted into an issue/request first, I’m happy to adapt the contribution format.

jperla:
Updated this PR on current `main` and addressed the Cursor review findings. Changes made:

- provider exclusions now include registered provider integrations, so `trustedrouter` can be excluded even if the current deployed-model snapshot has no TrustedRouter-bound row
- added a regression test for that exclusion-list behavior
- made the down migration rollback-safe by removing TrustedRouter BYOK rows before restoring the old provider check

Focused checks pass locally:

```sh
go test ./internal/providers ./internal/api/admin ./internal/proxy ./internal/translate ./internal/router/catalog
```

GitHub now reports the branch as mergeable from my side.

jperla:
Looks like the review bot skipped this since I'm not on the allowed-authors list — would appreciate a maintainer triggering a review when convenient. Happy to restructure the provider addition however you prefer; quick to iterate.

jperla:
Rebased onto current main. Resolved conflicts with the new anthropic_gateway provider (kept both providers in docs, provider maps, and the settings page), renumbered the migration to 0048 after the 0037 monthly-spend-limits clash, and added trustedrouter to the new RL roster coverage test (defaultRosterPrefixProviders — slash-form IDs need no vendor prefix, same as openrouter). Full `go test ./...` passes.

jperla:
Fixed both Bugbot findings and rebased onto current main.

The provider `CHECK` rewrite was built from a stale allowlist — it dropped `bedrock`, `makora`, `together`, `xai`, and `anthropic_gateway`. Both directions now carry 0045's set, and the down migration restores it exactly. Renumbered to 0050 after 0048/0049 landed.

On the catalog: added the missing TrustedRouter fallback for `moonshotai/kimi-k3`. Checking the rest of the bindings the same way turned up two more that would have 404'd on failover — `qwen/qwen3-coder` and `qwen/qwen3.5-flash-02-23` sent the catalog ID verbatim, but upstream they are `qwen/qwen3-coder-480b-a35b-instruct` (matching the Fireworks binding) and `qwen/qwen3.5-flash-2026-02-23`. Both now pin `UpstreamID`.

`go build ./...` and `go test ./...` pass; `go run ./cmd/genprices` output is unchanged.

jperla:
Rebased onto current `main`.

Two things moved under us and are followed here:

- The settings provider-key panel moved to `frontend/src/components/settings/ProviderKeysPanel.tsx`, so the TrustedRouter entry moved with it.
- `0050` is taken upstream, so the BYOK allowlist migration is renumbered `0060` and now carries the `0052` allowlist forward. The earlier version would have dropped `openai_gateway` on apply — that was Bugbot's "Migration drops valid providers" finding. The down migration deletes `trustedrouter` rows before restoring the constraint, which covers the "Down migration breaks with BYOK" one.

In the catalog rows where both sides added bindings, `trustedRouterBinding` trails OpenRouter and the Wafer bindings stay trailing-most, per the comment on those rows.

`go build ./...` and `go test ./internal/providers/... ./internal/router/... ./internal/translate/...` pass locally.

jperla:
Fixed both Bugbot findings and rebased onto current `main`.

**Duplicate migration version.** Renumbered `0060` → `0084`. `0060_installation-first-request-served` is upstream's and migrations now run through `0083`, so `golang-migrate` would have rejected the duplicate and migrate-up would have failed. The allowlist still carries `0052`'s set forward — I checked that `0061`, `0064` and `0067` touch this table but leave the provider `CHECK` alone, so `0052` is still the live constraint.

This is the fourth renumber on this branch (0048 → 0050 → 0060 → 0084). If it would help, I'm happy to switch the migration to a timestamp-style version instead, or to re-number on request right before merge.

**Missing V4-Pro binding.** Added the TrustedRouter binding to `deepseek/deepseek-v4-pro-0813`. It serves that id directly, so no `UpstreamID` pin is needed.

**One thing the review didn't flag, but is the same class of bug.** Every TrustedRouter binding on the branch carried the *OpenRouter* row's price rather than TrustedRouter's own, so all 16 were wrong — and since the scorer picks on price, the wrong ones route wrongly in both directions. `xiaomi/mimo-v2.5-pro` was listed at $1.00/$3.00 against an actual $0.422/$0.918, so it would never have been picked; `deepseek/deepseek-v4-flash` was listed at $0.14/$0.28 against an actual $0.072/$0.177. All 16 are now read from `GET /v1/models`, with `CacheReadMultiplier` derived from the published `input_cache_read` where there is one (`qwen/qwen3-235b-a22b-2507` publishes a cache read at parity with its input price, hence the explicit `1.0000` rather than falling through to the 0.5 default).

Prices are a snapshot; say the word if you would rather this branch carry no TrustedRouter prices at all and let them come from `genprices`.

Also carried the module rename (`workweave/router` → `weave-os/router`) through the files this branch adds.

`go build ./...` and the full `go test ./...` pass; `go run ./cmd/genprices` output is unchanged.

jperla:
Rebased onto current `main` (through #1386).

**Migration renumbered `0084` → `0099`.** Upstream `0084` is now `0084_session-pin-demoted-models`, and migrations run through `0098`, so the duplicate version would have failed `migrate up`. This is the fifth renumber on this branch — happy to switch to a timestamp version, or to renumber once more right before merge, whichever you prefer.

Re-checked that the allowlist the migration carries forward is still the live one: nothing after `0067` touches `model_router_external_api_keys`, so `0052`'s provider `CHECK` set is unchanged and the constraint is rewritten with exactly that set plus `trustedrouter`.

The only content conflict was the `docs/CONFIGURATION.md` env table — kept both rows (`ROUTER_MODEL_ID_MAP` and the two `TRUSTEDROUTER_*` entries). `go build ./...` and the provider/catalog/admin/translate test packages are green.

jperla:
Rebased again onto current `main` (through #1405).

**Migration renumbered `0099` → `0101`** — #1396 landed `0099_subscriber-prepaid-ownership` and migrations now run through `0100`. That is the sixth renumber on this branch, the second in two days. I am going to stop guessing at a free number: if you are willing, I will switch this to a timestamp-style version (e.g. `20260920_add-trustedrouter-provider`), which cannot collide with anything you merge in the meantime. Say the word and I will push it; otherwise I will keep renumbering on request.

The other conflict was `catalog.go`: #1397 added `Source: SourceOpenSource` to the `moonshotai/kimi-k3` entry, which our comment change collided with. Kept your `Source` classification and the TrustedRouter fallback comment together.

Re-verified the migration still carries the live allowlist — nothing after `0067` touches `model_router_external_api_keys`, so `0052`'s provider `CHECK` set is unchanged. `go build ./...` and the provider/catalog/admin test packages are green.

PR #525, opened 2026-06-26
feat(routing-report): add --per-prompt-json flag for replay scoring

steventohme:
## What

Adds an additive `--per-prompt-json <path>` flag to `cmd/routing-report`. When set, it writes the target version's per-probe routed pick as `[{id, model}]` in corpus order to the given path, alongside the normal markdown report.

## Why

This is the local-only dev-tool patch the replay scorer (in the sibling `router-internal/eval/` harness) depends on. Committing it removes the dependency of a CI gate on an uncommitted local patch.

## Behavior

- Additive only — no change to the Scorer, the markdown output, or any existing flag.
- The dump only fires when `--per-prompt-json` is set, reusing the already-computed target route (`tgt.chosen[i]` per corpus prompt).
- Reuses the file's existing `fatal()` error helper and the already-imported `json` / `os` packages.

## Test

```
go build -tags no_onnx ./cmd/routing-report/...   # OK
go vet   -tags no_onnx ./cmd/routing-report/...   # OK
gofmt -l cmd/routing-report/main.go               # clean
```

https://claude.ai/code/session_01JnJ7bzxbMVgYNPkeqgMAmB

### DIFF
--- a/cmd/routing-report/main.go
+++ b/cmd/routing-report/main.go
@@ -144,6 +144,7 @@ func main() {
 		topP         = flag.Int("top-p", 2, "top-P clusters blended per decision (prod runs 2)")
 		qualityBias  = flag.Float64("quality-bias", -1, "QualityBias dial position in [0,1] to route at; <0 uses the bundle's default knobs (no dial)")
 		outPath      = flag.String("out", "", "write markdown here instead of stdout")
+		perPromptOut = flag.String("per-prompt-json", "", "also dump [{id,model}] per probe (target chosen model, corpus order) here")
 	)
 	flag.Parse()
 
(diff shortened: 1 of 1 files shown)


caioribeiroclw-pixel:
Small schema suggestion while this is still additive: consider making the dump versioned and a little less model-only, e.g.

```json
{
  "schema_version": 1,
  "rows": [
    { "id": "...", "target_model": "..." }
  ]
}
```

If the replay scorer later needs to explain a routing delta, a top-level schema/version buys room to add fields such as `target_provider`, `decision_reason`, cache-policy bucket, or score metadata without changing the file shape from an array to an object later.

The `[{id, model}]` shape is totally fine for the immediate consumer; the reason I’d consider versioning now is that routing eval artifacts tend to become audit artifacts quickly once people ask “why did this prompt switch models?” or “did the cache/cost tradeoff change?”. A tiny envelope keeps this useful for replay while leaving space for those route-decision receipts later.

PR #565, opened 2026-07-01
Fix(Billing) :rotate preserve spend cap

### LINKED ISSUE #564: RotateAPIKey silently drops spend_cap_usd_micros on the replacement key — no API or UI path exists to restore it
## Summary

`RotateAPIKey` issues a replacement key for a revoked one, carrying forward
its name — but not its spend cap. The new key is always uncapped
(`spend_cap_usd_micros = NULL`), even when the key it replaces had a cap set.

Since there is no update/PATCH endpoint for API keys, and no dashboard field
for the spend cap at all, the cap can currently only be set via a direct
database write. This means the *only* way to restore a cap lost to rotation
is another manual out-of-band write — there is no user- or admin-facing
recovery path.

## Reproduction (verified locally)

1. Create a router API key (dashboard or `POST /admin/v1/keys`)
2. Set a spend cap directly in the DB:
```sql
   UPDATE router.model_router_api_keys
   SET spend_cap_usd_micros = 50000000  -- $50
   WHERE id = '<key id>';
```
3. Confirm the cap is set:
```sql
   SELECT id, spend_cap_usd_micros FROM router.model_router_api_keys WHERE id = '<key id>';
   -- spend_cap_usd_micros = 50000000
```
4. Rotate the key via the dashboard ("Rotate" button) or `POST /admin/v1/keys/:id/rotate`
5. Query the replacement key:
```sql
   SELECT id, spend_cap_usd_micros FROM router.model_router_api_keys
   WHERE name = '<key name>' ORDER BY created_at DESC LIMIT 1;
   -- spend_cap_usd_micros = NULL
```

Confirmed end-to-end with both the raw API and the dashboard's Rotate button —
same result either way.

## Root cause

- `auth.APIKey` and `auth.CreateAPIKeyParams` (`internal/auth/api_key.go`)
  have no `SpendCapUsdMicros` field
- `toAuthAPIKey` (`internal/postgres/converters.go`) drops
  `row.SpendCapUsdMicros` when converting the DB row to the domain type
- `CreateModelRouterAPIKey` (`db/queries/model_router_api_keys.sql`) does
  not insert `spend_cap_usd_micros`
- `RotateAPIKey` (`internal/auth/service.go`) reads the old key (which
  never had visibility into its own cap, per the above) and calls
  `IssueAPIKey` with no cap to carry forward

Enforcement itself (`internal/server/middleware/api_key_spend_cap.go`) is
correct — it reads live from Postgres and hard-rejects with 402 at the cap.
This is purely a gap in the create/rotate path, not an enforcement bug.

## Impact

Any capped key becomes silently uncapped the next time it's rotated for any
reason (including routine security hygiene unrelated to the cap). Neither
the admin performing the rotation nor anyone reviewing the dashboard
afterward has any way to notice — the dashboard doesn't surface
`spend_cap_usd_micros` anywhere, before or after rotation, and the rotate
confirmation dialog gives no indication that anything beyond the token
itself is affected.

## Proposed fix

- Add `SpendCapUsdMicros *int64` to `auth.APIKey` and `CreateAPIKeyParams`
- Propagate it in `toAuthAPIKey`
- Add `spend_cap_usd_micros` to the `CreateModelRouterAPIKey` INSERT
  (via `sqlc.narg` for nullability), regenerate with `sqlc generate`
- Pass it through in `apiKeyRepo.Create`
- In `RotateAPIKey`, carry the old key's cap forward to the new key
  (likely cleanest as a small `IssueAPIKeyWithCap` variant to avoid
  threading an optional param through every existing `IssueAPIKey` call site)

Image

rohith500:
Fixes #564
 
## Problem
 
`RotateAPIKey` issues a replacement key via `IssueAPIKey`, which never carried forward `spend_cap_usd_micros`. Any capped key silently became uncapped on rotation, with no warning to the caller and no dashboard signal — the admin key list doesn't surface the cap field at all, before or after rotation.
 
Since there's no update/PATCH endpoint for keys, and no dashboard field for the cap, the only way to set — or restore — a cap is a direct database write. That makes a dropped cap permanent without out-of-band intervention: an admin doing routine key rotation for unrelated reasons (security hygiene, a leaked token) silently removes a customer's spend limit with zero indication anything beyond the token itself changed.
 
## Root cause
 
- `auth.APIKey` and `CreateAPIKeyParams` (`internal/auth/api_key.go`) had no `SpendCapUsdMicros` field
- `toAuthAPIKey` (`internal/postgres/converters.go`) dropped `row.SpendCapUsdMicros` when converting the DB row to the domain type
- `CreateModelRouterAPIKey`'s INSERT (`db/queries/model_router_api_keys.sql`) never wrote `spend_cap_usd_micros`
- `RotateAPIKey` (`internal/auth/service.go`) read the old key through this same blind spot before calling `IssueAPIKey`, so there was nothing to carry forward even in principle
Enforcement itself (`internal/server/middleware/api_key_spend_cap.go`) is correct and unaffected by this bug — it reads `spend_cap_usd_micros` / `spent_usd_micros` live from Postgres and hard-rejects with 402 at the cap. This is purely a gap in the create/rotate path.
 
## Fix
 
- Add `SpendCapUsdMicros *int64` to `APIKey` / `CreateAPIKeyParams` (nil = uncapped, preserves existing behavior for every current uncapped key)
- Propagate it in `toAuthAPIKey`
- Extend the `CreateModelRouterAPIKey` INSERT with `spend_cap_usd_micros` via `sqlc.narg(...)` so it stays nullable/optional; regenerated with sqlc v1.30.0
- Thread it through `apiKeyRepo.Create`
- Add `IssueAPIKeyWithCap`; `IssueAPIKey` now delegates to it with a nil cap, so every existing call site (fresh key creation via `POST /admin/v1/keys`) is unaffected. `RotateAPIKey` calls `IssueAPIKeyWithCap` with the old key's cap.
## Files changed
 
| File | Change |
|---|---|
| `internal/auth/api_key.go` | Added `SpendCapUsdMicros *int64` to `CreateAPIKeyParams` |
| `internal/postgres/converters.go` | `toAuthAPIKey` now propagates `SpendCapUsdMicros` from the DB row |
| `db/queries/model_router_api_keys.sql` | `CreateModelRouterAPIKey` INSERT extended with `spend_cap_usd_micros` via `sqlc.narg(...)` |
| `internal/sqlc/model_router_api_keys.sql.go` | Regenerated (sqlc v1.30.0) — `CreateModelRouterAPIKeyParams` gains `SpendCapUsdMicros *int64`, query gains one param |
| `internal/postgres/repository.go` | `apiKeyRepo.Create` passes `params.SpendCapUsdMicros` through to sqlc params |
| `internal/auth/service.go` | Added `IssueAPIKeyWithCap`; `IssueAPIKey` delegates with nil cap; `RotateAPIKey` calls it with `target.SpendCapUsdMicros` |
 
Two commits, test-then-fix:
 
- `9b50052` — `test(auth): assert RotateAPIKey preserves spend cap` (failing test only)
- `3725463` — `fix(auth): carry spend cap forward on RotateAPIKey`
## Testing
 
**New test, before the fix:**
 
```
--- FAIL: TestRotateAPIKey_preservesSpendCap (0.00s)
    rotate_spend_cap_test.go:116:
        Error: Expected value not to be nil.
        Messages: spend cap must be carried forward to the replacement key;
                  got nil (cap was 5000000 micros on old key)
```
 
**Same test, after the fix:**
 
```
--- PASS: TestRotateAPIKey_preservesSpendCap (0.00s)
```
 
**Full suite:** `go test ./... -race -count=1` — all packages green, race detector clean, all 20 pre-existing auth tests still pass.
 
`make check` — clean.
 
**Manual end-to-end verification** — rebuilt the local server image (`docker compose build server && docker compose up -d server`) and ran the full flow through the real running stack, not just the unit test, to confirm the fix works through the actual admin API and DB.
 
Capped key, before rotation:
 
| id | spend_cap_usd_micros |
|---|---|
| `0b60fab5-e82a-4220-bba2-07b369aca241` | `5000000` |
 
`POST /admin/v1/keys/0b60fab5-.../rotate` → `201`, returns new key `f4727177-6d57-4cc4-948d-bc557b6fe14d`
 
Both rows after rotation:
 
| external_id | spend_cap_usd_micros | created_at | |
|---|---|---|---|
| `kid_xgVBu9eH7N2GqZpwQfYo3rRL` | `5000000` | 2026-07-01 19:12:30 | new key, cap preserved |
| `kid_Kdts2E1HObZEm638KlQk4hwV` | `5000000` | 2026-07-01 19:11:40 | old key |
 
Regression check — uncapped key through the same rotate path, confirming the fix doesn't introduce a cap where none existed:
 
| external_id | spend_cap_usd_micros | |
|---|---|---|
| `kid_0CZuqdy4J5dAGq70Djf9Rfgg` | `null` | new key, still uncapped |
| `kid_329GSrWAUoWTfiLQdAPIZupq` | `null` | old key |
 
## Also verified via the dashboard UI
 
Reproduced the original bug and confirmed the fix through the actual "Rotate" button in the admin dashboard (not just curl), since that's the real-world trigger path. Screenshots in #564.

### DIFF
--- a/db/queries/model_router_api_keys.sql
+++ b/db/queries/model_router_api_keys.sql
@@ -6,7 +6,8 @@ INSERT INTO router.model_router_api_keys (
     key_prefix,
     key_hash,
     key_suffix,
-    created_by
+    created_by,
+    spend_cap_usd_micros
 )
 VALUES (
     @installation_id::uuid,
@@ -15,7 +16,8 @@ VALUES (
     @key_prefix::varchar,
     @key_hash::varchar,
     @key_suffix::varchar,
-    @created_by
+    @created_by,
+    sqlc.narg('spend_cap_usd_micros')
 )
 RETURNING *;
 
--- a/internal/auth/api_key.go
+++ b/internal/auth/api_key.go
@@ -6,27 +6,29 @@ import (
 )
 
 type APIKey struct {
-	ID             string
-	InstallationID string
-	ExternalID     string
-	Name           *string
-	KeyPrefix      string
-	KeyHash        string
-	KeySuffix      string
-	LastUsedAt     *time.Time
-	CreatedAt      time.Time
-	DeletedAt      *time.Time
-	CreatedBy      *string
+	ID                string
+	InstallationID    string
+	ExternalID        string
+	Name              *string
+	KeyPrefix         string
+	KeyHash           string
+	KeySuffix         string
+	LastUsedAt        *time.Time
+	CreatedAt         time.Time
+	DeletedAt         *time.Time
+	CreatedBy         *string
+	SpendCapUsdMicros *int64
 }
 
 type CreateAPIKeyParams struct {
-	InstallationID string
-	ExternalID     string
-	Name           *string
-	KeyPrefix      string
-	KeyHash        string
-	KeySuffix      string
-	CreatedBy      *string
+	InstallationID    string
+	ExternalID        string
+	Name              *string
+	KeyPrefix         string
+	KeyHash           string
+	KeySuffix         string
+	CreatedBy         *string
+	SpendCapUsdMicros *int64
 }
 
 type APIKeyRepository interface {
--- a/internal/auth/rotate_spend_cap_test.go
+++ b/internal/auth/rotate_spend_cap_test.go
@@ -1,0 +1,120 @@
+package auth_test
+
+// TestRotateAPIKey_preservesSpendCap is a FAILING test that documents the bug:
+// RotateAPIKey issues the replacement key via IssueAPIKey, which calls Create
+// without forwarding spend_cap_usd_micros. The new key always gets a nil cap,
+// silently removing any budget limit that was set on the original key.
+//
+// Fix: propagate SpendCapUsdMicros through CreateAPIKeyParams so RotateAPIKey
+// can carry the old key's cap onto the new one.
+
+import (
+	"context"
+	"testing"
+
+	"workweave/router/internal/auth"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// rotateCapKeyRepo is a minimal APIKeyRepository that supports the Create and
+// ListForInstallation calls that RotateAPIKey drives. Create stores the new key
+// keyed by ID; ListForInstallation returns whichever keys belong to that
+// installation. SoftDelete marks a key deleted but leaves it visible only to
+// ListForInstallation (so RotateAPIKey can find the old key to copy from).
+type rotateCapKeyRepo struct {
+	keys []*auth.APIKey
+	next int
+}
+
+func (r *rotateCapKeyRepo) Create(_ context.Context, p auth.CreateAPIKeyParams) (*auth.APIKey, error) {
+	r.next++
+	key := &auth.APIKey{
+		ID:             auth.GenerateID("kid"),
+		InstallationID: p.InstallationID,
+		ExternalID:     p.ExternalID,
+		Name:           p.Name,
+		KeyPrefix:      p.KeyPrefix,
+		KeyHash:        p.KeyHash,
+		KeySuffix:      p.KeySuffix,
+		CreatedBy:      p.CreatedBy,
+		// SpendCapUsdMicros intentionally NOT set — this mirrors the real
+		// CreateModelRouterAPIKey SQL which has no spend_cap_usd_micros column
+		// in its INSERT, so new keys always start with a nil cap.
+		SpendCapUsdMicros: nil,
+	}
+	r.keys = append(r.keys, key)
+	return key, nil
+}
+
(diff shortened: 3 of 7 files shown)


rohith500:
@aminsamir45 

pushed a rebase onto current main (5716e85) — should be ready for the workflows to run whenever you get a chance to approve them.
 
FYI on the earlier failure (the "Routing review" run against `3725463`): it wasn't related to this PR's changes. That run was diffed against `c3c0977` as its base commit, and the routing-report tool couldn't find `centroids.bin` for the v0.71 artifact bundle at that point in history (`open centroids.bin: no such file or directory`). Confirmed the file exists fine on current `main` (49KB, retrievable via `git show`), so it looks like it was just a stale/broken baseline at the time this branch was first opened — should be resolved now that it's rebased onto a newer `main`.

rohith500 (commented):
@aminsamir45

pushed a rebase onto current main to pick up a routing artifact fix that was causing the earlier CI failure (unrelated to this PR's changes). Ready for the required workflows to be approved whenever you get a chance. 

Thanks!

PR #570, opened 2026-07-02
fix(usage): treat Anthropic 429 as subscription exhaustion so failover engages

steventohme:
## Problem

Subscription customers hit their personal Claude plan cap and then eat **raw 429s on every subsequent turn** instead of failing over — even when the deploy has an `ANTHROPIC_API_KEY` to serve them on. Observed on Everfit (`org_ytbWhf0XxuuKLR3XAjtviSl0`): **37 subscription-429s in 24h, zero failovers.**

The exhaustion-failover chain (`claudeSubscriptionExhausted` → `withSuppressedSubscription` → deployment/BYOK key) was fully wired and the fallback key was present. The **trigger** was the gap: `Snapshot.Exhausted()` is learned only from `anthropic-ratelimit-unified-*` utilization headers, but the Claude OAuth **session-limit 429 is a bare `rate_limit_error` carrying none of those headers**:

```
{"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your account's rate limit. Please try again later."}}
```

So exhaustion was never recorded, the spent OAuth token got re-injected every turn, and the failover never engaged.

## Fix

Detect exhaustion from the **429 status itself**, in addition to the header path:

- `UpstreamHeaderObserver` / `ObserveUpstreamHeaders` now carry the response **status code** (both provider callers — anthropic, openai — updated).
- On a 429 for the caller's Claude subscription token, `withUsageObserver` records exhaustion via new `Observer.RecordExhausted` / `RecordExhaustedFromHeaders`, with retention **bounded by `Retry-After` / `*-reset`** so suppression lifts at the real refill (else a ~5h fallback window, so a transient cap self-heals rather than staying suppressed for a week).

The next turn sees the plan as spent → suppresses the subscription → serves on the deployment/BYOK Anthropic key (billed full cost). Self-heals in-flight on the bypass→routed fallthrough. Unchanged when no fallback key exists (subscription is kept rather than 400ing on no credential).

## Caveats

- The **one** turn that first trips the cap still 429s once — that's the learning signal; it self-heals from the next turn.
- Failover serves Claude on the **Weave key at full price** — real platform spend on subscription orgs, but strictly better than a hard failure.

## Tests

- `usage/exhaustion_test.go` — `RecordExhausted` marks exhausted; header-less 429 outlives the short ttl but expires by the 5h window; `Retry-After` and absolute `unified-*-reset` pin retention to the real reset; the weekly window is preserved (not erased) when only the session cap 429s.
- `subscription_exhaustion_429_internal_test.go` — a subscription 429 flips `claudeSubscriptionExhausted` on the next turn (with a fallback key); stays disengaged with no fallback key; a 200 at 20% used is not misread as exhaustion.

`go vet ./...` clean, full suite green.


steventohme:
Closing as redundant. #555 (fail over on 429 on the subscription path) already fixes this root cause, and does it better — it rescues the throttled turn in-flight (same-turn retry on the Weave/BYOK key), so the user never sees a 429, whereas this PR only suppresses the spent sub on the *next* turn. #570 was built against a stale gitlink pin that predated #555 landing on main; the deployed code has since caught up. No correctness gap remains.

PR #578, opened 2026-07-02
Treat client history trims as free-switch windows and add cold-pin follow-fresh lever

aminsamir45:
## Summary

Tier 1 of the inter-turn routing optimizations. Two changes:

**Prefix-trim free-switch** (on by default, kill switch `ROUTER_PREFIX_TRIM_FREE_SWITCH`):

- Compaction/trim detection (`compactionTracker.checkAndRecord`) moves from the post-routing block in `ProxyMessages` into `runTurnLoop`, before the pin/planner path, so the signal is available on the exact turn the client broke its prompt prefix.
- On a detected trim, the planner prices the pin's upstream cache as **cold** even inside the provider TTL — the client rewrote the transcript head, so cache entries keyed on the old prefix are unreachable no matter how recently the pin served. This flips "warm-cache inertia" stays into switches exactly when switching is free.
- The switch handover summarizer is skipped on trim turns: the body already carries the client's own compaction summary and is as small as it will ever be; summarizing it again is pure cost.
- The post-routing non-Anthropic compaction handover now consumes the turn loop's verdict (`routeRes.PrefixTrimmed`) instead of re-recording the tracker (re-recording would compare the turn's counts against themselves and never fire).

**Cold-pin follow-fresh** (off by default, `ROUTER_SWITCH_COLD_PIN_FOLLOW_FRESH`): a new planner lever that overturns an EV stay when the pin's cache is cold — with nothing warm to preserve, staying buys no cache reuse, so the scorer's fresh argmax wins. EV-positive and tier-upgrade keep precedence. 

## Test plan

- [x] `go build ./...`, `go vet ./...`, full `go test ./...` green
- [x] New planner unit test: cold+lever → switch (`cold_pin_follow_fresh`), lever-off → stay, warm → stay, EV-positive precedence
- [x] New turn-loop integration tests: trim turn flips a warm EV-negative stay into a switch with the summarizer skipped; `WithPrefixTrimFreeSwitch(false)` preserves the warm stay; a sub-agent's small opening turn does not inherit the main loop's trim baseline

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -645,7 +645,12 @@ func main() {
 		ThresholdUSD:           parseEnvFloat("ROUTER_SWITCH_EV_THRESHOLD_USD", proxy.DefaultPlannerThresholdUSD),
 		ExpectedRemainingTurns: parseEnvInt("ROUTER_SWITCH_EXPECTED_REMAINING_TURNS", proxy.DefaultPlannerExpectedRemainingTurns),
 		TierUpgradeEnabled:     config.GetOr("ROUTER_SWITCH_TIER_UPGRADE_ENABLED", boolDefault(proxy.DefaultPlannerTierUpgradeEnabled)) == "true",
+		ColdPinFollowFresh:     config.GetOr("ROUTER_SWITCH_COLD_PIN_FOLLOW_FRESH", boolDefault(proxy.DefaultPlannerColdPinFollowFresh)) == "true",
 	}
+	// Prefix-trim free-switch: on a detected client history trim, price the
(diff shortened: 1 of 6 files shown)


aminsamir45:
* **#578** <a href="https://app.graphite.com/github/pr/workweave/router/578?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/578?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #681, opened 2026-07-09
fix(proxy): expire hmm_history and guard cyber-refusal repin from HMM override

rohith500:
## Summary
 
Fixes a session-routing bug where a cyber-refusal fallback pin is silently
undone on the very next HMM turn.
 
1. **Expire `{role}_hmm_history` on cyber-refusal repin** — closes the gap
   Bugbot flagged on #675 ("Refusal repin ignores sticky role"). #675 fixed
   the call site to use `stickyStateRole(routeRes)`, but `maybeRepinOnRefusal`
   never expired `hmm_history`, so it kept pointing at the refusing model.
2. **Treat `cyber-refusal-repin` as a valid HMM stay candidate** (via
   `isHMMPinReason`) so the fallback pin itself can actually win EV-stay,
   instead of being discarded as a non-HMM pin.
3. **Block `hmm_confident_upgrade` while a `cyber-refusal-repin` pin is
   active** — found via live e2e repro, not speculative. The mock HMM
   sidecar returned a high-confidence pick for the refusing model, which
   would have silently defeated commits 1+2 in production. Scoped to the
   pin's *reason*, not the specific model (see tradeoff below).
## Root cause
 
`recordHMMTurnHistory` writes the refusing model into `{role}_hmm_history`
before `maybeRepinOnRefusal` runs. `isHMMPinReason` excludes
`cyber-refusal-repin`, so on the next HMM turn `hmmStayPin` discards the
active fallback pin and falls back to `hmm_history` — which still points at
the refusing model with a live TTL. HMM EV-stay re-selects the refusing
model, undoing the re-pin.
 
Unaffected by #677 (confirmed — its diff touches `PriorServedModel`/
confidence-threshold plumbing only, not `maybeRepinOnRefusal`, `hmmStayPin`,
or history expiry).
 
## Live verification
 
End-to-end on a rebuilt container from this branch (mock HMM sidecar,
`ROUTER_CYBER_REFUSAL_REPIN=true`):
 
- **Turn 1** (HMM session, refusal triggered): `claude-opus-4-8` refuses.
  Server re-pins to fallback.
```
  session_pins:
    default_mid              → claude-sonnet-5   (cyber-refusal-repin, active)
    default_mid_hmm_history  → claude-opus-4-8    (expired)
```
 
- **Turn 2** (same session):
```
  X-Router-Decision: cyber-refusal-repin
  X-Router-Model: claude-sonnet-5
```
 
  `decision_reason: cyber-refusal-repin`, `sticky_hit: true` — fallback holds.
 
Prior to commit 3, the same repro showed turn 2 reverting to `claude-opus-4-8`
via `hmm_confident_upgrade` even after commits 1–2.
 
## Tradeoff
 
The confident-upgrade block in commit 3 is scoped to the active pin's
*reason*, not to the specific refusing model — the refusing model isn't
persisted on the active pin row (`UpsertSessionPin` omits `last_served_model`),
so narrowing this further would require a schema/field addition. Net effect:
no confident upgrade to *any* model (including a genuinely better third
option) while the repin is active. Acceptable tradeoff for short-term
stability after a refusal; the pin expires on its normal TTL.
 
## Test plan
 
- [x] `TestHMMEVStay_RespectsCyberRefusalRepin` — fails on `main`, passes here
- [x] `TestHMMEVStay_CyberRefusalRepinBlocksConfidentUpgrade`
- [x] `TestHMMCostGate_IgnoresNonHMMActivePin` — unchanged, still passes
      (confirms the carve-out is scoped to `cyber-refusal-repin` only, not a
      general relaxation of the non-HMM pin gate)
- [x] `make check`
- [x] `go test ./... -race -count=1` — 3 pre-existing failures
      (`TestFireTelemetryRecoversFromPanic`, `TestSafeGoRecoversFromPanic`,
      `TestService_VerifyAPIKey_RecoversFromMarkUsedPanic`), confirmed
      reproducing identically on `main`, unrelated to this change
- [x] Live e2e repro (see above)

### DIFF
--- a/internal/proxy/cyber_refusal_internal_test.go
+++ b/internal/proxy/cyber_refusal_internal_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 
@@ -41,6 +42,15 @@ func (f *repinFakeStore) ResetUpstreamErrors(context.Context, [sessionpin.Sessio
 	return nil
 }
 func (f *repinFakeStore) SweepExpired(context.Context) error { return nil }
+
+func activeCyberRefusalUpsert(upserts []sessionpin.Pin) (sessionpin.Pin, bool) {
+	for _, u := range upserts {
+		if u.Reason == "cyber-refusal-repin" && u.PinnedUntil.After(time.Now()) {
+			return u, true
+		}
+	}
+	return sessionpin.Pin{}, false
+}
 
 // A realistic Anthropic-native refusal SSE (the router-visible wire shape:
 // stop_reason "refusal" on HTTP 200; see catalog.go). The real opus cyber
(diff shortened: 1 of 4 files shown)


rohith500:
@steventohme

this closes out the hmm_history half of the gap
your Bugbot review flagged on #675 ("Refusal repin ignores sticky role").
The stickyStateRole fix landed there, but hmm_history itself never got
expired, so cyber-refusal's fallback pin was getting silently overridden
on the next HMM turn. Confirmed live end-to-end (details in the PR
description).

PR #713, opened 2026-07-14
fix(ci): use pull_request_target so comment review runs on forks

steventohme:
## Summary
Switch the comment-length review workflow from `pull_request` to `pull_request_target` so it runs on forked PRs, not just same-repo branches.

`pull_request` events from forks don't receive repo secrets or OIDC tokens, so `anthropics/claude-code-action` failed with "Could not fetch an OIDC token." `pull_request_target` runs in the base repo's context with secrets available.

## Safety
The review is advisory-only (COMMENT event, not REQUEST_CHANGES) and the Claude step only runs `git diff` + `gh api` — no user-authored build scripts execute with the bot token. The checkout uses `ref: ${{ github.event.pull_request.head.sha }}` so the diff is against the PR's code.

## Changes
- `pull_request` → `pull_request_target`
- Checkout `ref: head.sha` so the diff inspects the PR branch, not base
- Explicit `git diff ...head.sha` instead of `...HEAD` (ambiguous in `pull_request_target` context)
- Removed fork guard `if:` condition (no longer needed)

### DIFF
--- a/.github/workflows/comment_length_review.yml
+++ b/.github/workflows/comment_length_review.yml
@@ -5,11 +5,12 @@ name: Comment Length Review
 # the shortened comment. Never blocks merge (no required status check, and the
 # review is posted as event=COMMENT, not REQUEST_CHANGES).
 #
-# Comments are authored by the workweave-bot account (WEAVE_BOT_TOKEN PAT), not
-# github-actions[bot]: the gh api reviews call below runs under that token.
+# Runs on pull_request_target so forked PRs also get reviewed. Safe because the
+# review is advisory-only (COMMENT event, no write to the fork's code) and only
+# runs Claude on the git diff, not user-authored build scripts.
 
 on:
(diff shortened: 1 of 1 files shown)


rohith500:
@steventohme 

This fixes a real gap that's been blocking review on Devin's PRs. 

Since those come in from forked branches, the comment_length_review.yml workflow was triggering on pull_request, which doesn't pass repo secrets or OIDC tokens to forked runs. Result: anthropics/claude-code-action fails immediately with "Could not fetch an OIDC token," so none of the recent Devin PRs have gotten a comment-length review at all.
Switching to pull_request_target runs the workflow in the base repo's context, so secrets/OIDC are available regardless of where the branch lives. To keep this safe on a privileged trigger:


Would like to get this merged soon since it's currently a silent gap in CI coverage — Devin PRs are merging without ever having gone through the comment-length check.

PR #714, opened 2026-07-14
fix: separate router liveness from HMM readiness

steventohme:
## Problem

Cloud Run revisions can receive HMM-routed traffic before the optional policy sidecar is ready. Coupling that dependency to `/health` would also turn a sidecar outage into a router liveness failure, even while the default cluster strategy remains healthy.

## Fix

Add a dedicated `/readyz` endpoint that checks the HMM sidecar through the existing authenticated policy client. `/health` remains process-only liveness, and readiness failures return a stable public response without internal URLs.

## Changes

- reuse the Google ID-token-capable policy client for `/readyz`
- keep `/health` independent of optional sidecars
- remove the variadic checker API that silently dropped extra checkers
- exclude readiness polling from APM traces
- add handler, route, status, and auth regression tests

## Deployment

Keep the Cloud Run liveness probe on `/health`. Point the startup probe to `/readyz` when the updated router pin is consumed.

## Validation

- `wv mr tc`
- `wv mr t`


steventohme:
* **#714** <a href="https://app.graphite.com/github/pr/workweave/router/714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #772, opened 2026-07-17
fix(router): forward PriorServedModel in recordTurnUsage to latch has_ever_switched

### LINKED ISSUE #751: session-pin has_ever_switched latch never fires on the regular (non-HMM) writeback path
## TL;DR
 
`recordTurnUsage` (the normal per-turn writeback) forgets to pass one field (`PriorServedModel`) that `recordHMMTurnHistory` (the HMM writeback) already passes correctly. That field is what lets a brand-new session-pin row learn "this session already switched models" from its `_hmm_history` sibling. Miss it, and once that sibling row expires, the session's stale-signed Anthropic `thinking` blocks stop getting stripped — reopening the exact 400 class #383 fixed. One-line fix, verified live end-to-end below (bug reproduced, fix confirmed, both with actual captured request bytes).
 
## Summary
 
`internal/proxy/service.go` has two session-pin usage-writeback functions: `recordTurnUsage` (line 3149, the regular path) and `recordHMMTurnHistory` (line 3185, the HMM path). #677 added a `PriorServedModel` field so a brand new tier-role pin row could inherit switch history from its `<role>_hmm_history` sibling row. `recordHMMTurnHistory` sets it (line 3218). `recordTurnUsage` never does — its `sessionpin.Usage{}` literal (lines 3164-3171) just doesn't have the field. These are the only two places in the repo that construct a `sessionpin.Usage{}`, so this is the complete picture, not a sample.
 
**Net effect:** any tier-role pin row whose first-ever write happens on the regular (non-HMM) path can never latch `has_ever_switched = true` from prior evidence — not "sometimes," structurally can't, because the one clause that could catch it depends on a value that's never bound. See "Exact mechanism" below.
 
**Why you won't have seen this in normal testing:** it's masked at read time by a fallback OR (`hmmHistory.HasEverSwitched`) for as long as the `_hmm_history` sibling row is alive. It only becomes observable after that sibling expires — see "Why it's latent" below.
 
**Fix is one line** — see "Fix" section. Everything else here is the evidence trail.
 
## Exact mechanism
 
`db/queries/session_pins.sql:102-114` (`UpdateSessionPinUsage`):
 
```sql
UPDATE router.session_pins
SET last_input_tokens        = @last_input_tokens::int,
    last_cached_read_tokens  = @last_cached_read_tokens::int,
    last_cached_write_tokens = @last_cached_write_tokens::int,
    last_output_tokens       = @last_output_tokens::int,
    last_turn_ended_at       = @last_turn_ended_at::timestamptz,
    has_ever_switched        = has_ever_switched
      OR (last_served_model <> '' AND last_served_model <> @last_served_model::varchar)
      OR (@prior_served_model::varchar <> '' AND @prior_served_model::varchar <> @last_served_model::varchar),
    last_served_model        = @last_served_model::varchar
WHERE session_key = @session_key::bytea
  AND role        = @role::varchar;
```
 
On a row's first-ever write, `last_served_model` (the pre-update value) is `''`, so the second OR-clause is false by construction — nothing to compare against yet. The third clause is the only thing that can still catch a real prior switch on that fresh row, and it only fires if `@prior_served_model` is non-empty. `recordTurnUsage` never binds it, so on that path the third clause is always `'' <> '' AND ...` which is always false.
 
The signal exists, it's just discarded. `res.PriorServedModel` is computed every turn regardless of HMM classification, by `switchHistoryFromPins(pin, hmmHistory)` at `turnloop.go:340`, where `hmmHistory` comes from `loadHMMHistory` (`turnloop.go:1007`) — which reads role `<role>_hmm_history` (`hmmHistoryRole`, `turnloop.go:134`) unconditionally, not just on HMM turns. By the time either writeback function runs, the correct value is sitting right there in `res`. `recordTurnUsage` just doesn't use it.
 
## Why it's latent
 
`switchHistoryFromPins` (`turnloop.go:1020-1033`) also runs at read time, every turn:
 
```go
sessionEverSwitched := activePin.HasEverSwitched || hmmHistory.HasEverSwitched
```
 
While `<role>_hmm_history` is alive, this masks the bug completely. It only surfaces once that sibling is gone:
 
- Both rows share the same sliding TTL (`pinExpiry`, `service.go:257-261`), refreshed only by whichever function writes to that specific row. A session where HMM fires early then goes quiet, while regular tier turns keep going, drifts the two rows' expiry apart, and `_hmm_history` ages out first.
- `SweepExpired` (`session_pins.sql:139-145`) deletes it 24h later.
- After that, `hmmHistory.HasEverSwitched` reads `false` (zero-value row), and the tier row's own never-latched `false` becomes the only source of truth for every subsequent turn on that role.
## Reproduction (verified live, exact steps)
 
Local docker stack, mock HMM sidecar, mock Anthropic upstream, installation `5bf0b947-5c55-49b6-b9e7-b346cba19193`, role `default_low`.
 
1. HMM turn switches `default_low` from `claude-haiku-4-5` (A) to `claude-fable-5` (B). Row `default_low_hmm_history`: `last_served_model = 'claude-fable-5'`, `has_ever_switched = true`. No `default_low` row exists yet.
2. Regular (non-HMM) turn — first write to `default_low` itself, served by A again. Row `default_low`: `last_served_model = 'claude-haiku-4-5'`, `has_ever_switched = false` — this is the bug. Confirmed via psql: `@prior_served_model` bound on this write was empty.
3. Delete/expire `default_low_hmm_history` (simulating TTL drift + sweep).
4. Same-model turn on `default_low` (still A), client transcript carries an assistant `thinking` block signed for B — realistic, since Claude Code resends full history every turn. Captured outbound body to mock Anthropic:
```json
[
  { "type": "thinking",
    "thinking": "prior-model-B chain of thought that should be stripped",
    "signature": "stale-sig-from-claude-fable-5" },
  { "type": "text", "text": "visible assistant text from prior turn" }
]
```
 
Stale-signed block reached the wire unstripped. Real Anthropic would reject this with 400 "Invalid signature in thinking block."
 
5. Counterfactual: patched `recordTurnUsage` locally to forward `PriorServedModel`, repeated steps 2-4 identically. Step 2 now shows `has_ever_switched = true`. Step 4's captured body:
```json
[ { "type": "text", "text": "visible assistant text from prior turn" } ]
```
 
Block correctly stripped.
 
Trace for step 4: `envelope.go:1029` (`if opts.ModelSwitched { ov.StripThinkingBlocks = true }`) never fires because `opts.ModelSwitched` comes from `service.go:2336` (`ModelSwitched: routeRes.modelSwitched()`, confirmed this is the `/v1/messages` path exercised, not the OpenAI path at line 4235), which comes from `SessionEverSwitched`, which comes from the never-latched column.
 
## Ruling out a race
 
`recordTurnUsage` and `recordHMMTurnHistory` can't both fire for one turn:
 
```go
if isHMMTurn(res) {
    s.recordHMMTurnHistory(res, ...)
    return
}
```
 
`isHMMTurn` is a pure function of the routing decision, not timing. This is a one-sided omission, not a concurrency bug.
 
## Fix
 
`internal/proxy/service.go`, `recordTurnUsage`, lines 3164-3171:
 
```go
usage := sessionpin.Usage{
    InputTokens:       in,
    CachedReadTokens:  cacheRead,
    CachedWriteTokens: cacheCreation,
    OutputTokens:      out,
    EndedAt:           time.Now(),
    ServedModel:       servedModel,
    PriorServedModel:  res.PriorServedModel,   // add this
}
```
 
## Acceptance criteria
 
- New test: given a pin store where `<role>_hmm_history` already shows `HasEverSwitched=true` and a different `LastServedModel`, a `recordTurnUsage` call that's the first write to `<role>` persists `has_ever_switched=true` on that row (the stored column, not just the live-read value).
- No behavior change for non-HMM sessions (`hmmHistory` is always zero-value there, so `PriorServedModel` is `""` regardless).
- Existing `recordHMMTurnHistory` coverage unaffected.
## Out of scope (flagging only)
 
- No `ANTHROPIC_BASE_URL`-equivalent override exists in prod wiring for local Anthropic-path mocking (Google/OpenAI have one) — patched `main.go` locally to verify this, reverted before filing.
- Cross-tier switches (`default_low` to `default_high`) get fully separate rows with no shared bridge at all — a broader design question predating #677, not a regression, out of scope here.

steventohme:
## Problem

`recordTurnUsage` (the regular, non-HMM session-pin writeback) never sets `PriorServedModel` in its `sessionpin.Usage{}` literal. The SQL latch in `UpdateSessionPinUsage` needs `@prior_served_model` to set `has_ever_switched = true` on a fresh pin row (one with no `last_served_model` yet).

Without it, a brand-new tier-role pin row can never inherit switch history from its `<role>_hmm_history` sibling:

1. HMM turn writes `_hmm_history` with `has_ever_switched=true`
2. A non-HMM turn creates the tier-role pin — `has_ever_switched` stays `false` because `@prior_served_model` is empty
3. `_hmm_history` expires (TTL drift between the two rows)
4. Now `has_ever_switched=false` everywhere — stale thinking blocks from the prior model are not stripped
5. A subsequent Anthropic turn gets `400: thinking blocks in messages must have a signature`
6. Session is unrecoverable

## Fix

One line in `recordTurnUsage` (`internal/proxy/service.go`):

```
PriorServedModel: res.PriorServedModel,
```

`res.PriorServedModel` is already computed every turn by `switchHistoryFromPins` — it just was not being forwarded to the SQL write.

New test `TestRecordTurnUsage_ForwardsPriorServedModel` asserts the field is passed through to `UpdateUsage`.

Closes #751

### DIFF
--- a/db/queries/session_pins.sql
+++ b/db/queries/session_pins.sql
@@ -93,12 +93,10 @@ ON CONFLICT (session_key, role) DO UPDATE SET
 -- /force-model upsert cannot overwrite the genuinely-last-served model
 -- before the next turn reads it to detect a mid-session model switch.
 -- has_ever_switched latches true the first time the just-served model
--- differs from a prior non-empty last_served_model, including caller-supplied
--- prior-served evidence when the stored row is new. ModelSwitched (derived from
--- last_served_model) strips stale Anthropic thinking-block signatures only on
--- the single transition turn, but Claude Code re-sends its full transcript
--- every turn, so once a session has switched at all, the latch keeps the emit
--- path stripping on every subsequent same-model turn for the session's life.
+-- differs from a prior non-empty last_served_model. Caller-supplied model and
+-- latch evidence preserves history when the stored role row is new.
+-- The latch keeps stripping stale thinking signatures on later turns because
+-- clients resend the full transcript.
 -- name: UpdateSessionPinUsage :exec
 UPDATE router.session_pins
 SET last_input_tokens        = @last_input_tokens::int,
@@ -107,6 +105,7 @@ SET last_input_tokens        = @last_input_tokens::int,
     last_output_tokens       = @last_output_tokens::int,
     last_turn_ended_at       = @last_turn_ended_at::timestamptz,
     has_ever_switched        = has_ever_switched
+      OR @session_ever_switched::boolean
       OR (last_served_model <> '' AND last_served_model <> @last_served_model::varchar)
       OR (@prior_served_model::varchar <> '' AND @prior_served_model::varchar <> @last_served_model::varchar),
     last_served_model        = @last_served_model::varchar
--- a/internal/postgres/session_pin_repo.go
+++ b/internal/postgres/session_pin_repo.go
@@ -74,6 +74,7 @@ func (r *SessionPinRepo) UpdateUsage(ctx context.Context, sessionKey [sessionpin
 		LastTurnEndedAt:       pgtype.Timestamptz{Time: endedAt.UTC(), Valid: true},
 		LastServedModel:       usage.ServedModel,
 		PriorServedModel:      usage.PriorServedModel,
+		SessionEverSwitched:   usage.SessionEverSwitched,
 	})
 }
 
--- a/internal/proxy/failover_integration_test.go
+++ b/internal/proxy/failover_integration_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"workweave/router/internal/providers"
+	"workweave/router/internal/providers/openai"
 	"workweave/router/internal/providers/openaicompat"
 	"workweave/router/internal/proxy"
 	"workweave/router/internal/router"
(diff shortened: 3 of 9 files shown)


steventohme:
* **#772** <a href="https://app.graphite.com/github/pr/workweave/router/772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/772?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #836, opened 2026-07-25
feat(catalog): reconcile gpt-5.6-sol/terra/luna onto v0.75 (v0.76-v0.78)

aminsamir45:
## Summary

`gpt-5.6-sol/terra/luna` were catalog-added in #721 (2026-07-15), but their roster-onboarding overlay (`add-gpt56-v075-overlay`, commit `4ae87e1`, "v0.75 = v0.74 + gpt-5.6-*") was never merged. A different `v0.75` — `v0.75 = v0.73 + claude-opus-5` (#833) — forked from an earlier parent and landed first, winning the version name and orphaning the gpt-5.6 branch. Net effect: Sol/Terra/Luna never shipped to `deployed_models`, so the Weave dashboard never showed them despite the catalog/pricing/force-model work being live for over a week.

This PR re-appends all three onto the **current** (post-opus-5) `v0.75` as three sequential column-append versions:

- `v0.76` = v0.75 + `gpt-5.6-sol` (TierHigh, $5/$30)
- `v0.77` = v0.76 + `gpt-5.6-terra` (TierHigh, $2.5/$15)
- `v0.78` = v0.77 + `gpt-5.6-luna` (TierMid, $1/$6)

$0 spend — reuses Tier-3 SWE-bench Easy shards already sitting on GCS from the original onboarding probes (`add-gpt-5.6-{sol,terra,luna}`, 582 shards each, 2026-07-15/16). No new inference.

## Two gotchas fixed along the way

1. **Parity-gate mismatch**: `append_quality_column.py`'s `load_version_knobs()` reads the per-cluster alpha-floor *vector* from `metadata.yaml`'s `training.default_routing_knobs` — but this lineage's `rankings.json` was actually baked with a **uniform alpha=0.96** blend (per `build_v075_opus5_swap.py`, proven against v0.73). The vector reading fails the tool's own parity gate (Δ=0.159); forcing uniform alpha=0.96 reproduces v0.73/v0.75 bit-for-bit (Δ=0.0). This affects every version in the v0.68/v0.70–v0.75 lineage, not just this PR.
2. **Impute-only starvation**: fitting each model's column against only its own shards starves the ≥4-incumbent-per-cluster minimum the fit needs, silently falling back to pure imputation everywhere. Pulled the combined Tier-3 Easy corpus from 7 sibling onboarding runs already on GCS (fable-5, sonnet-5, kimi-k2.7, minimax-m3, qwen3.7-plus, glm-5.2) so coding clusters 5/8/13/14 get real fitted cells instead.

Full writeup: `router-internal/docs/eval/RESULTS_gpt56_v076_v078_reconcile_opus5_fork_20260724.md` (companion doc lives in the WorkWeave monorepo, not this repo).

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all 58 packages pass
- [x] Parity gate (Δ=0.0 vs. immediate parent's `rankings.json`) at each of the three append steps
- [x] `offline_screen.py --versions v0.75,v0.76,v0.77,v0.78` — no coding-cluster argmax regression; new models enter as viable non-winning targets at current evidence level (same pattern as prior additive onboardings)
- [ ] **NOT done, and intentionally out of scope for this PR**: live staging bake-off. Per `ROUTER_MODEL_LIFECYCLE.md`, `artifacts/latest` must not move without one — this PR does not touch `artifacts/latest` and should not be promoted without a follow-up bake-off PR.


aminsamir45:
Closing: wrong surface.

Prod and staging run `ROUTER_DEFAULT_STRATEGY=hmm` (`terraform/environments/workweave-{prod,staging}-01/main.tf`), so the cluster artifacts this PR builds do not decide production routing — they serve only installations explicitly pinned to the `cluster` strategy.

gpt-5.6 sol/terra/luna have in fact been routable in prod since 2026-07-21: they are already arms in the pinned HMM bandit state (`hmm-gpt56-sol-complete-499-20260721`) — luna+terra in `balanced`, sol+terra in `high`, sol in `maximum`. So v0.76–v0.78 would have changed nothing that matters.

Not lost work — two findings from this build are being carried into the docs/skills so the next person doesn't repeat it:
1. **Parity-gate gotcha:** `append_quality_column.py`'s `load_version_knobs()` reads the per-cluster alpha-floor vector from `metadata.yaml`, but this lineage's `rankings.json` was baked with uniform alpha=0.96. The vector reading fails the tool's own parity gate (Δ≈0.16) on v0.68 and v0.70–v0.75; forcing uniform alpha reproduces the parent bit-for-bit.
2. **Fit starvation:** passing only the new model's own Tier-3 shards starves `--min-fit-models` (≥4) so every cluster silently imputes. The sibling-run corpus from `mine_bakeoff_priors.DEFAULT_RUNS["easy"]` is needed for a real fit.

If we ever need gpt-5.6 on the `cluster` strategy specifically, this branch is a working starting point.

PR #852, opened 2026-07-29
feat(proxy): warn Claude Code when a subscription turn fails over to the billable Weave key

steventohme:
## Why

Investigating an anomalous router-usage session ($836 on one org's Weave API key), we traced the proximate cause to: a Claude subscription hit its 5-hour unified rate limit mid-session (`anthropic-ratelimit-unified-5h-status: rejected`), and the router correctly failed the turn over to the billable Weave/BYOK Anthropic key — both credential-suppression paths already handle this — but did so silently. The customer had no signal that subsequent turns would be billed instead of served on their own plan.

## What

Adds `subscriptionFailoverWarningMarker`, a new routing-marker constant distinct from `subscriptionOnlyWarningMarker`:
- `subscriptionOnlyWarningMarker` — permanent, org-level depleted-credits state, with a top-up CTA.
- `subscriptionFailoverWarningMarker` (new) — per-turn, self-healing condition ("this turn", not "until you add credits"), no CTA, not gated by the routing-marker opt-out (same billing-state-change rule as the existing marker).

Wired into both places a subscription-served Anthropic turn can fail over:
1. **Pre-emptive** (`service.go` ~2457) — a prior observer snapshot already reads exhausted before dispatch.
2. **Reactive** (`service.go` ~2886) — a live 429/OAuth-rejection triggers the subscription-credit retry onto the Weave key. Swap is safe because this branch only runs pre-commit (`preludeBuf.Committed() == false`).

## Out of scope

Codex/OpenAI: there's no `codexSubscriptionExhausted` equivalent gating that path today, so a Codex subscription failover stays silent until that symmetric guard is built (separate follow-up).

## Tests

Two new full-flow `ProxyMessages` integration tests in `failover_integration_test.go` — one per path — asserting the marker text appears and the retry authenticates via the deployment key, not the spent subscription bearer. Full suite: `wv mr tc && wv mr t` pass.

### DIFF
--- a/internal/proxy/failover_integration_test.go
+++ b/internal/proxy/failover_integration_test.go
@@ -550,7 +550,8 @@ func TestProxyMessages_PreemptiveSubscriptionFailoverWarnsUser(t *testing.T) {
 	obs.Record(obs.Key([]byte(bypassSubToken)), usage.Snapshot{
 		Secondary: usage.Window{UsedPercent: 1.0, WindowMinutes: 10080},
 	})
-	svc := proxy.NewService(fr, map[string]providers.Client{providers.ProviderAnthropic: p}, nil, false, nil, nil, false, providers.ProviderAnthropic, bypassScorerPickMdl, nil).
+	telemetry := newCaptureTelemetry()
+	svc := proxy.NewService(fr, map[string]providers.Client{providers.ProviderAnthropic: p}, nil, false, nil, nil, false, providers.ProviderAnthropic, bypassScorerPickMdl, telemetry).
 		WithSubscriptionAwareRouting(obs, 0.05, 2.0).
 		WithDeploymentKeyedProviders(map[string]struct{}{providers.ProviderAnthropic: {}})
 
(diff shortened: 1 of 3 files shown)


steventohme:
Pushed a second commit adding `credential.source` and `dispatch.subscription_failover_billable` to both Anthropic and OpenAI upstream span builders. This closes a gap needed for a follow-up WorkWeave-side admin alert: `credential_source` previously only reached this router's own Postgres telemetry insert, never the OTel span/log attrs that `recordCallLog` reuses as the `router.call` metadata base -- so WorkWeave's backend ingest (which consumes that OTLP log record) had no way to see it. `dispatch.subscription_failover_billable` combines both the pre-emptive and reactive failover paths since the existing `dispatch.subscription_failover` only covers the reactive one.

steventohme:
Depended on by workweave/WorkWeave#11247 (subscription-failover admin alert) — that PR reads `credential.source` / `dispatch.subscription_failover_billable` from the OTLP log record this PR adds them to.

PR #861, opened 2026-07-31
fix(translate): strip unsigned thinking blocks on every Anthropic emit

### LINKED ISSUE #860: fix(proxy): HMM pin-session TTL gap silently strips switch history, re-opening the #383 thinking-signature 400 class
## Incident

Two of the largest prod installations (b253074f, b56c551) are experiencing 400s with "Invalid `signature` in `thinking` block" against `claude-opus-5`:

- Session `3a74d50a-619d-4bef-922a-ca7fc10c3ee7`: 57 of 152 opus-5 turns 400'd on 2026-07-30
- The 400s are **buffered** (preludeBuf not committed) and the error body is verbatim:
  `{"error":{"type":"invalid_request_error","message":"messages.7.content.0: Invalid \`signature\` in \`thinking\` block"}}`
- All turns are `ProxyMessages` (native Anthropic wire), `cross_format=false`, subscription OAuth, HMM strategy
- Zero failovers: 4xx buffered body is deliberately non-retryable, but nothing retries it

## Root cause: pin-session TTL gap → lost switch history → unsigned thinking blocks flow through

**The timeline:**

1. Jul 21-22: session serves `claude-opus-4-8` (Anthropic, same provider). Thinking blocks are **signed** (native Anthropic signatures).
2. Jul 27-29: session serves non-Anthropic models (kimi-k2.7, minimax-m3, deepseek-v4-flash) in `cross_format=true` turns. The router's translation layer converts these models' responses into **unsigned** thinking blocks — the Anthropic-format blocks carry *no* `signature` field because OSS providers don't emit them.
3. Jul 29 11:02: HMM decides `claude-opus-5`. The transition turn calls `modelSwitched()` → `PriorServedModel=minimax-m3, Decision.Model=claude-opus-5` → `transition=true` → **`StripThinkingBlocks=true`**. Success.
4. The `modelSwitched()` result is critical: it strips **all** historical thinking blocks, signed and unsigned alike. This cleanses the transcript of the unsigned cross-format blocks.
5. **HMM history row and regular pin row both have a 1-hour TTL** (`pinSessionTTL = time.Hour`, `service.go:255`). Both are refreshed every turn via `recordHMMTurnHistory` and `recordTurnUsage`.
6. Session goes idle from ~11:09 Jul 29 to ~06:50 Jul 30 (~19 hours). **Both rows expire during the gap.**
7. Jul 30 06:50: first turn after the gap. `switchHistoryFromPins(pin, hmmHistory)` returns `("", false)` — both rows are fresh, `LastServedModel=""`, `HasEverSwitched=false`.
8. `modelSwitched()` returns `false` (no prior model, no switch evidence). **`StripThinkingBlocks` stays false.**
9. The unsigned thinking blocks from step 2, still in Claude Code's transcript, are forwarded verbatim to opus-5.
10. Anthropic rejects with `Invalid signature in thinking block` → 400.

**The next turns compound the problem:** the 400s produce no `thinking` output blocks, so `recordHMMTurnHistory` writes `PriorServedModel=<as-before>` with no model change. Suffix of 400s continues producing more 400s until the session times out or the client gives up.

## Why #383's fix alone isn't enough

#383 added `ModelSwitched → StripThinkingBlocks` (line 1114 of `envelope.go`), and the turn-loop correctly computes `modelSwitched()` (line 161 of `turnloop.go`). This works perfectly for **contiguous sessions** (no gap longer than 1 hour).

But the guard depends entirely on the pin store to carry switch history. The pin-session TTL is deliberately 1 hour (matches Anthropic prompt-cache TTL). A session that goes idle past that window loses all switch history, and there is no fallback path to detect cross-format thinking blocks in the body before forwarding them.

## Fix options

**Option A — inspect the body for unsigned thinking blocks (defensive):** In `resolveAnthropicOverrides`, add a lightweight scan after the existing `ModelSwitched` check: if `StripThinkingBlocks` is still false but the body contains a `"thinking"` content block **without** a `"signature"` field, set `StripThinkingBlocks = true`. This is a pure defensive check that doesn't require any persisted state.

```go
// Defensive: unsigned thinking blocks from an earlier cross-model turn
// surface after the pin's switch-history TTL expires. Scan the content
// array for any thinking block missing its signature field.
if !ov.StripThinkingBlocks {
    ov.StripThinkingBlocks = bodyHasUnsignedThinking(body)
}
```

A `bodyHasUnsignedThinking` helper scans `messages[].content[]` for entries where `type == "thinking"` and `signature` is absent, stopping at the first hit. The scan is O(bytes) on the request body but bounded (largest bodies are ~1MB in 329 messages).

**Option B — extend TTL for switch history:** Make the pin rows carry switch history beyond the 1-hour window. More invasive: touches DB schema, risk of pin-store bloat, and doesn't help existing unanchored sessions.

**Option C — rewrite the thinking blocks differently at translation time:** Produce `reasoning_preview` (a carrier Claude Code already handles) instead of unsigned `thinking` blocks on the Anthropic emit path for OSS models. Solves the problem at the source but is a larger translate-surface change.

**Recommendation: Option A** — it's the smallest defensive fix, doesn't touch the DB, fully covers the gap independent of pin-TTL, and is consistent with the existing belt-and-suspenders approach in the translate layer (we already defensively strip thought_signature and sanitize tool IDs on Anthropic targets). Option C could complement it as follow-up.

## Impact

- 2 installations affected, ~62 errors in 4 days
- All are hard failures: client sees a 400 and must restart the session
- The error is silent from the router's perspective — `fireTelemetry` records it as `upstream_status=400` but there's no special-case handling for `Invalid signature` class errors

## Related

- #383 — original `ModelSwitched → StripThinkingBlocks` fix
- #751 — the `PriorServedModel` writeback gap (FIXED; deployed)
- #752 — the fix for #751 in the regular (non-HMM) writeback path (FIXED; deployed)

steventohme:
Fixes #860.

## Problem (live in prod)

Two of the largest installations are hitting repeated 400s against `claude-opus-5`:

```
messages.7.content.0: Invalid `signature` in `thinking` block
```

One session (`3a74d50a…`) lost **57 of 152 turns** on 2026-07-30. The 400 is buffered, but 4xx is deliberately non-retryable, so nothing rescues the turn — the user sees a hard failure and has to restart.

## Root cause

A session served OSS models (kimi-k2.7, minimax-m3, deepseek-v4-flash) for several turns. Cross-format emit converts their reasoning into Anthropic `thinking` blocks with **no `signature`** — OSS providers have none to carry. Claude Code keeps those blocks in its transcript and replays them on every subsequent turn.

#383 already strips carried-over thinking blocks, but only when `ModelSwitched` is set — and that is derived entirely from the session pin's switch history. Those rows carry a one-hour TTL (`pinSessionTTL`, `service.go:255`). The session idled ~19h; both the pin and its `_hmm_history` sibling lapsed; `switchHistoryFromPins` returned `("", false)`; `ModelSwitched` was false. The unsigned blocks went upstream verbatim.

So the existing guard is correct but not sufficient: it protects contiguous sessions and silently stops protecting anything that idles past the pin TTL.

## Fix

Strip unconditionally on Anthropic emit, scoped to blocks whose `signature` is absent or empty (`StripUnsignedThinkingBlocks`, a floor under the existing `StripThinkingBlocks`).

Signature-scoped rather than blanket on purpose: a signed history stays intact when no switch is known, so prompt-cache continuity and reasoning context are preserved — only the shapes Anthropic *always* rejects are dropped. `redacted_thinking` is exempt; it legitimately carries `data` instead of a signature.

This needs no persisted state, so it holds regardless of pin TTL, idle gaps, or which writeback path ran.

## Verification

- Reproduced the exact prod payload shape (unsigned block at `messages[7].content[0]`, `ModelSwitched=false`, target `claude-opus-5`): **fails before the change, passes after** — confirmed by reverting just the override assignment and re-running.
- `go build ./...`, `go vet ./...` clean; full `go test ./...` green (46 packages).

New tests: unsigned stripped without a model switch; empty-signature stripped; **signed block survives alongside an unsigned one being dropped** (the reason for the signature scoping); `redacted_thinking` untouched.

## One existing test changed

`TestAnthropicSameFormat_ThinkingBlocksKeptForCapableModel`'s fixture had carried an unsigned thinking block since #1. That shape cannot originate from Anthropic (real thinking blocks always carry a signature), so the fixture now includes one. The assertion — capable targets keep their thinking history — is unchanged and still passes.

## Follow-up (not in this PR)

Worth considering separately: emitting a different carrier (e.g. `reasoning_preview`) for OSS reasoning on the Anthropic path, so unsigned `thinking` blocks are never synthesized at all. This PR is the minimal stop-the-bleeding fix.

### DIFF
--- a/internal/translate/emit_same_format_test.go
+++ b/internal/translate/emit_same_format_test.go
@@ -732,6 +732,26 @@ func TestAnthropicSameFormat_SignedThinkingSurvivesUnsignedStrip(t *testing.T) {
 	assert.Equal(t, "text", text["type"])
 }
 
+// TestAnthropicSameFormat_UnsignedThinkingOnlyMessageDropped is the exact shape
+// flagged in #861 review: an assistant replay whose ONLY content is an unsigned
+// thinking block (no accompanying text/tool_use). Stripping the block must drop
+// the whole message rather than emit content:[], which Anthropic also rejects.
+func TestAnthropicSameFormat_UnsignedThinkingOnlyMessageDropped(t *testing.T) {
+	body := []byte(`{"model":"claude-opus-4-7","messages":[{"role":"user","content":"hi"},{"role":"assistant","content":[{"type":"thinking","thinking":"from an OSS model"}]},{"role":"user","content":"continue"}],"max_tokens":1024,"thinking":{"type":"adaptive"}}`)
+	opts := translate.EmitOptions{
+		TargetModel:   "claude-opus-4-7",
+		Capabilities:  router.Lookup("claude-opus-4-7"),
+		ModelSwitched: false,
+	}
+	out := parseAndEmit(t, body, "anthropic", opts)
+	msgs, _ := out["messages"].([]any)
+	require.Len(t, msgs, 2, "the unsigned-thinking-only assistant message must be dropped entirely")
+	first, _ := msgs[0].(map[string]any)
+	assert.Equal(t, "user", first["role"])
+	second, _ := msgs[1].(map[string]any)
+	assert.Equal(t, "user", second["role"])
+}
+
 func TestPassthroughSameFormat_FieldsScrubbed(t *testing.T) {
 	body := []byte(`{"model":"claude-sonnet-4-20250514","messages":[{"role":"user","content":"hi"}],"max_tokens":1024,"effort":"high","thinking":{"type":"enabled"},"context_management":{"mode":"auto"},"output_config":{"length":"verbose"}}`)
 	env, err := translate.ParseAnthropic(body)
@@ -883,6 +903,9 @@ func TestAnthropicSameFormat_ManyThinkingBlocksInSingleMessage(t *testing.T) {
 	assert.Equal(t, "final", content[0].(map[string]any)["text"])
 }
 
+// TestAnthropicSameFormat_AllThinkingBlocksRemoved guards the Greptile P1 on
+// #861: a message whose content becomes empty after stripping must be dropped
+// entirely, not left behind as content:[] — Anthropic rejects that shape.
 func TestAnthropicSameFormat_AllThinkingBlocksRemoved(t *testing.T) {
 	body := []byte(`{"model":"claude-sonnet-4-20250514","messages":[{"role":"user","content":"hi"},{"role":"assistant","content":[{"type":"thinking","thinking":"t1"},{"type":"redacted_thinking","data":"xyz"}]}],"max_tokens":1024}`)
 	opts := translate.EmitOptions{
@@ -891,10 +914,9 @@ func TestAnthropicSameFormat_AllThinkingBlocksRemoved(t *testing.T) {
 	}
 	out := parseAndEmit(t, body, "anthropic", opts)
 	msgs, _ := out["messages"].([]any)
-	require.Len(t, msgs, 2)
-	assistantMsg, _ := msgs[1].(map[string]any)
-	content, _ := assistantMsg["content"].([]any)
-	assert.Len(t, content, 0, "all blocks are thinking blocks, content should be empty array")
+	require.Len(t, msgs, 1, "the assistant message must be dropped, not left with an empty content array")
+	userMsg, _ := msgs[0].(map[string]any)
+	assert.Equal(t, "user", userMsg["role"])
 }
 
 func TestAnthropicSameFormat_StringContentPreserved(t *testing.T) {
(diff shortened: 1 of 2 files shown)


steventohme:
Fixed in 77cbe621 — `rewriteMessageBlocks` now drops a message entirely when stripping empties its content array, instead of leaving `content:[]` behind. This is fixed in the shared helper, so it also protects the existing `StripThinkingBlocks`/`ModelSwitched` path, not just the new unsigned-strip path. Added `TestAnthropicSameFormat_UnsignedThinkingOnlyMessageDropped` covering the exact unsigned-thinking-only shape from your replay artifact, and updated `TestAnthropicSameFormat_AllThinkingBlocksRemoved` (which had asserted the old `content:[]` behavior).

PR #863, opened 2026-07-31
feat(byok): support customer provider keys in managed mode with a platform fee

steventohme:
## Summary

BYOK was reachable only from the self-hosted admin dashboard. Managed deployments stripped stored BYOK keys at the auth boundary because a stored row would spend upstream *and* debit prepaid credits — so a hosted customer had no way to plug in their own upstream provider.

This PR opens that path. An opt-in per installation, billing a 5% platform fee on the customer's upstream spend instead of full inference cost.

## Changes

- **Migration `0044_managed-byok`**: widens the provider CHECK (it excluded `makora` — blocks the literal ask), adds `base_url` to external keys, adds `byok_enabled` to installations, adds `byok_fee` to the billing ledger's entry_type CHECK.
- **Per-installation BYOK gate**: `byokDisabled` (per-deployment) becomes `byokRequiresOptIn` (per-installation). An org opting in can't change credentialing for other tenants. Self-hosted ignores the flag entirely.
- **Per-request base URL**: adds `BaseURL` to `proxy.Credentials`, populated from the BYOK row, resolved at the 9 URL-construction sites across the 4 provider clients. Reuses the existing per-request credential-resolution seam — `openai/client.go` already swaps base URL per credential (for Codex), so this extends an existing pattern rather than inventing one.
- **5% fee**: `DebitForInference` writes two ledger rows in one statement when `ByokServed` — an `inference` row at delta 0 with notional cost preserved (so the savings dashboard still prices the turn), plus a `byok_fee` row for the actual debit. Override and subscription still outrank BYOK. Spend counters include the fee, else a capped key would never trip its cap on BYOK traffic.
- Balance gate is unchanged: BYOK turns must NOT be exempt, since the fee needs a funded balance.

## Verification

- Full `go test ./...` green.
- `go vet ./...` clean.
- Migration up → down → up against a local Postgres reproduces the existing schema (constraint rebuilds, no dangling rows).
- New tests: per-installation BYOK gating (managed-strips when not opted-in, managed-keeps when opted-in, self-hosted always); 5% fee math (exact `$100 upstream → $5 fee`; override and subscription outrank BYOK); provider-AAD rejection on mismatched provider.
- Smoke suite (path-gated, requires cassette) not run locally — should be exercised in CI before merge.

## Follow-up WorkWeave change

There is a paired PR on the WorkWeave side that depends on this landing: `weaverouter/crypto.go` writes BYOK secrets with the same Tink keyset and AAD derivation, and the Tink version (`v2.2.0`) is now an explicit direct dependency in WorkWeave's `go.mod`. **The gitlink pin in WorkWeave must be bumped to whatever squash-commits this PR produces before the WorkWeave PR can merge** — `TestRouterSchemaCompatibility` will PREPARE every statement in `backend/internal/app/weaverouter/postgres.go` and fail loudly against the current pinned commit if the new columns aren't there yet.

## Risks

- **AAD divergence between router and WorkWeave**: a future refactor that drift the two encryptors produces rows the router silently cannot decrypt at request time, surfacing as an upstream 401 rather than a write error. The WorkWeave side has a frozen ciphertext fixture catching this; the router side has a unit test that round-trips Encrypt→Decrypt internally.
- **Custom base URLs** (added at write boundary, covered by tests) steer router traffic. SSRF-adjacent surface — any regression that loosens validation will be caught in CI but worth a security review.
- **The env-var shadow guard in the self-hosted admin handlers** (`internal/api/admin/keys.go` 409s when the provider's env var is set) is unchanged and continues to shadow BYOK on self-hosted. Managed mode bypasses it entirely by writing through the WorkWeave control plane, not the dashboard. This is intentional per the fee model but worth confirming in review.

## Out of scope

- Anthropic BYOK: subscription pass-through still wins, so an Anthropic BYOK key would not get exercised by this change against Claude/Codex subscription customers. Worth a follow-up if anyone asks for it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -205,7 +205,7 @@ If new helper doesn't fit, justify new package in code comment before creating.
 `ROUTER_DEPLOYMENT_MODE` read at boot in `cmd/router/main.go`:
 
 - **`selfhosted`** (default): full dashboard at `/ui/*`, `/admin/v1/*` API (auth, metrics, keys, provider-keys, config, excluded-models), dashboard cookie auth all mounted. Provider keys read from env vars; missing keys keep providers registered for client-passthrough but exclude from hard-pin resolution.
-- **`managed`**: dashboard + `/admin/v1/*` not mounted at all — Weave-managed deploys have separate control plane. Every provider registered with empty deployment key; proxy service in BYOK-only mode, so request without BYOK or client-supplied auth for chosen provider 400s rather than silently spending platform budget. Setting variable to any other value panics at boot.
+- **`managed`**: dashboard + `/admin/v1/*` not mounted at all — Weave-managed deploys have separate control plane. Every provider registered with empty deployment key; proxy service in BYOK-only mode, so request without BYOK or client-supplied auth for chosen provider 400s rather than silently spending platform budget. Setting variable to any other value panics at boot. BYOK is **opt-in per installation** here (`byok_enabled` on the installation row, set by the control plane): until it's set, `withAPIKey` strips external keys so a stored key can't spend against a credit-billed deployment. An opted-in BYOK turn debits no inference cost and is charged a platform fee (`billing.ByokFeeRate`) as a separate `byok_fee` ledger row.
 
 When adding new endpoint, put inside `selfhosted` block in `server.Register` unless part of product surface (`/v1/*`, `/v1beta/*`, `/health`, `/validate`). Do not re-expose admin surface in managed mode.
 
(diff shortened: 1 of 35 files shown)


steventohme:
* **#863** <a href="https://app.graphite.com/github/pr/workweave/router/863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/863?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #900, opened 2026-08-12
feat(providers): remap outbound model IDs for deployment-key endpoints

### LINKED ISSUE #526: model alias support for provider variations
Some providers like opencode go have slightly different model names. For these cases, it'd be nice if the hard coded names can be remapped to the provider-specific name through an env var or similar config

https://opencode.ai/zen/go/v1/models

```
deepseek/deepseek-v4-flash -> deepseek-v4-flash
deepseek/deepseek-v4-pro   -> deepseek-v4-pro
moonshotai/kimi-k2.6       -> kimi-k2.6
xiaomi/mimo-v2.5-pro       -> mimo-v2.5-pro
z-ai/glm-5.2               -> glm-5.2
```

Bedrock is another example where it has proprietary names (I know this is already supported however it may be behind an openai compatible proxy and not directly accessible--we're using original Bedrock "runtime" behind an OpenAI-compatible proxy)

### LINKED ISSUE #541: Reproduction: model alias mismatch with OpenCode Go (related to #526)
## Summary

This is a concrete reproduction of the issue described in #526 — model ID mismatch when using a custom OpenAI-compatible upstream (OpenCode Go) via `OPENROUTER_BASE_URL`.

The router sends slash-form model IDs (`deepseek/deepseek-v4-flash`) in the request body, but OpenCode Go expects bare names (`deepseek-v4-flash`). There is no env-var mechanism to remap model names at runtime.

### 1. Start the router

```bash
docker compose up -d
```

### 2. Point OpenRouter at OpenCode Go

Add to `.env` (docker compose auto-reads this):

```
OPENROUTER_API_KEY=sk-your-opencode-go-key
OPENROUTER_BASE_URL=https://opencode.ai/zen/go/v1
```

Then recreate:

```bash
docker compose up -d --force-recreate router-server
```

### 3. Create a router API key

Login at `http://localhost:8080/ui/login` (password: `admin`), then create a key via Settings → API Keys.

### 4. Send a test request

```bash
curl -s -N "http://localhost:8080/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -H "x-api-key: rk_YOUR_KEY" \
  -d '{"model":"deepseek/deepseek-v4-flash","messages":[{"role":"user","content":"say hello in one word"}]}'
```

### 5. Actual result

```json
{"type":"error","error":{"type":"ModelError","message":"Model deepseek/deepseek-v4-pro is not supported"}}
```

The error format (`{"type":"error",...}`) confirms the request reached OpenCode Go. It was rejected because the model name sent was the router's public slash-form name instead of OpenCode Go's bare name.

## Root cause analysis

The OpenRouter client (`openaicompat.Client`) sends the body's `"model"` field verbatim — no `modelIDMap` rewriting happens for OpenRouter:

| Provider | Has modelIDMap? | Constructor in main.go |
|---|---|---|
| OpenRouter | **No** (public slug = upstream ID) | `NewClient(key, url)` |
| DeepInfra | Yes | `NewClientWithModelIDMap(key, url, upstreamIDsForProvider("deepinfra"))` |
| Fireworks | Yes | `NewClientWithModelIDMap(key, url, upstreamIDsForProvider("fireworks"))` |
| Bedrock | Yes | `NewClientWithModelIDMap(key, url, upstreamIDsForProvider("bedrock"))` |

The `modelIDMap` rewrites the body's `"model"` field at request time in `rewriteModelField()` (`internal/providers/openaicompat/client.go:119`). For OpenRouter this is a no-op because the map is nil.

OpenCode Go's `/v1/models` returns bare names (`deepseek-v4-flash`, `deepseek-v4-pro`, `kimi-k2.6`, etc.) but the router catalog uses slash-form IDs (`deepseek/deepseek-v4-flash`, `moonshotai/kimi-k2.6`, `z-ai/glm-5.2`, etc.).

The existing `rewriteModelField` infrastructure is correct — it just needs an env-var hook so self-hosters can supply runtime mappings.

## Proposed fix direction

An env var (e.g. `ROUTER_MODEL_ID_MAP`) — comma-separated `from=to` pairs — parsed at boot and merged into the OpenRouter client's modelIDMap. Example:

```bash
ROUTER_MODEL_ID_MAP="deepseek/deepseek-v4-flash=deepseek-v4-flash,deepseek/deepseek-v4-pro=deepseek-v4-pro,moonshotai/kimi-k2.6=kimi-k2.6,xiaomi/mimo-v2.5-pro=mimo-v2.5-pro,z-ai/glm-5.2=glm-5.2"
```

## Environment

- Router: main branch, docker compose stack
- Upstream: OpenCode Go (`https://opencode.ai/zen/go/v1`)

vignesh-chaturvedi:
## What

Adds `ROUTER_<PROVIDER>_MODEL_ALIASES`, a JSON map of catalog model ID → the model name a provider's endpoint publishes, applied to the outbound request body on the deployment-key path.

Closes #526
Closes #541

## Why

Pointing a provider at an OpenAI-compatible endpoint that isn't the vendor itself is already a supported, documented setup — `.env.example` describes `OPENROUTER_BASE_URL` as "override for vLLM/Together/etc." But if that endpoint publishes the catalog's models under different names, there is no way to rewrite them, and every turn 400s.

#541 reproduces it end to end: the router sends `deepseek/deepseek-v4-flash`, the gateway expects `deepseek-v4-flash`, and the response is `Model ... is not supported`.

Rewriting exists, but only on paths that don't cover this case:

| Path | Mechanism | Covers a re-named deployment endpoint? |
|---|---|---|
| Catalog `UpstreamID` | `upstreamIDsForProvider`, baked into the binding | No — fixed at build time, and OpenRouter/XAI/Meta bindings carry none |
| BYOK key (#892) | `ExternalAPIKey.ModelAliases` → `proxy.ApplyModelAlias` | No — per-key, and only non-empty on BYOK credentials |

So a self-hosted deployment using `OPENROUTER_API_KEY` + `OPENROUTER_BASE_URL` — exactly #541's repro — has no mechanism at all. This adds the third layer.

## How

`resolveModelAliases` reads `ROUTER_<PROVIDER>_MODEL_ALIASES` for each OpenAI-compatible provider and layers the entries over that provider's catalog-derived `UpstreamID` map. The merged map goes to the existing `NewClientWithModelIDMap`, so the rewrite itself reuses `rewriteModelField` unchanged — no new code on the request path.

Precedence, outermost wins: **catalog `UpstreamID` < `ROUTER_<PROVIDER>_MODEL_ALIASES` < BYOK `model_aliases`**. The BYOK layer keeps winning for free, because `Client.Proxy` already applies it after the catalog map.

Only the outbound wire name changes. Routing, pricing, and analytics stay keyed on the catalog ID.

**Failure modes.** A malformed value aborts boot (`panic`, consistent with `config.MustGet` and `ROUTER_DEPLOYMENT_MODE`) — the error names the env var. An alias naming a model outside the deployed catalog is logged and skipped rather than fatal, so retiring a model can't turn a stale alias into a failed start.

**Validation is against catalog IDs, not per-provider bindings**, deliberately. An operator pointing `OPENROUTER_BASE_URL` at a different gateway may well be served models the catalog binds elsewhere — #526's own list includes `z-ai/glm-5.2`, which the catalog binds to Together and Fireworks. Rejecting those would reject the use case.

**Placement.** The logic sits in `cmd/router` rather than `internal/config` because it needs `internal/router/catalog`, and the root guide requires `internal/config` to stay a leaf that imports nothing else under `internal/`. It is boot-time wiring, which is the composition root's job, and `cmd/CLAUDE.md` lists small env parsers as belonging there.

## Behavior change

None without configuration. `resolveModelAliases` with no env vars set produces exactly the maps in use today:

| Provider | Catalog-derived entries | Resolved entries |
|---|---|---|
| openrouter | 0 | 0 |
| fireworks | 16 | 16 |
| makora | 2 | 2 |
| minimax | 2 | 2 |
| together | 7 | 7 |
| xai | 0 | 0 |
| meta | 0 | 0 |
| wafer | 3 | 3 |
| bedrock | 4 | 4 |

OpenRouter, XAI and Meta now construct via `NewClientWithModelIDMap` where they previously called `NewClient`. All three resolve to zero entries, so the map is nil and `rewriteModelField` returns the body untouched — an unconfigured deployment puts the same bytes on the wire as before.

The first-party OpenAI client and Wafer's Anthropic-compatible surface keep their own catalog maps and are deliberately out of scope; this PR covers the nine OpenAI-compatible deployment-keyed providers.

## Testing

`make build`, `vet`, `test`, `test-statusline`, `test-install`, `inference-boundary`, `check-docs` and `check-agent-guides` are all green. No `db/queries/` or migration files are touched, so `internal/sqlc/` is untouched (I couldn't run `sqlc generate` locally).

**`model_aliases_wire_test.go` reproduces #541 at the wire.** An `httptest` stub stands in for a gateway publishing bare names, and the assertion is the `model` field it actually receives:

| Case | Model on the wire |
|---|---|
| No alias configured (the #541 bug) | `deepseek/deepseek-v4-flash` — what the gateway rejects |
| `ROUTER_OPENROUTER_MODEL_ALIASES` set | `deepseek-v4-flash` — what it accepts |
| A different, unaliased model | `xiaomi/mimo-v2.5-pro`, unchanged |

I chose this over booting the compose stack because it pins the exact byte #541 is about, runs in CI on every change, and needs no upstream key. It also closes a gap: nothing else asserts `modelIDMap` rewriting at all.

**`model_aliases_test.go`** adds parsing cases (empty, valid, malformed JSON, JSON array, empty key, empty value), env override application, override beating a catalog `UpstreamID`, unaliased bindings surviving the merge, unknown models skipped without error, and a malformed value erroring with the env var named. They discover a suitable model from `catalog.Models` at run time rather than pinning an ID, so catalog churn doesn't break them.

Mutation-checked: dropping the `merged[id] = upstreamID` write fails `TestResolveModelAliasesAppliesEnvOverride`, `TestResolveModelAliasesOverridesCatalogUpstreamID` and the wire test's alias case, while the two no-alias cases correctly keep passing.

**Not run:** the full `docker compose` stack against a real re-named endpoint. The evidence above is the test suite plus the resolved-map parity table.

## Docs

- `docs/CONFIGURATION.md` — new "Deployment-level model aliases" subsection plus a row in the provider table.
- `.env.example` — commented `ROUTER_OPENROUTER_MODEL_ALIASES` example next to the other provider vars.
- `cmd/CLAUDE.md` — `resolveModelAliases` added to the composition-root helper list. `cmd/AGENTS.md` is regenerated via `make generate-agent-guides`, and `make check-agent-guides` passes.

## Incidental

`upstreamIDsForProvider` moves into the new file alongside its callers. Its doc comment had drifted onto `registerDeploymentKeyedProvider` in `main.go`; moving the function reunites the two.

## Rebase note

Rebased onto current `main` after the move to `weave-os/router` — the original branch was cut about 290 commits back. Carried forward in the rebase: the three OpenAI-compatible providers added since (MiniMax, Meta, Wafer) are aliasable too, imports moved to the `weave-os/router` module path, `AGENTS.md` is now generated so only `cmd/CLAUDE.md` is hand-edited, and two test helpers were renamed to avoid colliding with `flag_registry_test.go`'s `discardLogger` and the new `internal/dispatch` package.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,6 +58,14 @@ POSTGRES_SSLMODE=require       # set to `disable` for local Docker
 # GOOGLE_API_KEY=...
 # GOOGLE_BASE_URL=https://generativelanguage.googleapis.com/v1beta/openai
 
+# Remap outbound model IDs per provider, for an OpenAI-compatible endpoint
+# that publishes the catalog's models under its own names. JSON map of
+# catalog model ID -> the name that endpoint expects. <PROVIDER> is one of
+# OPENROUTER, FIREWORKS, MAKORA, MINIMAX, TOGETHER, XAI, META, WAFER, or
+# BEDROCK. Routing, pricing, and analytics stay keyed on the catalog ID.
+# See docs/CONFIGURATION.md -> Deployment-level model aliases.
+# ROUTER_OPENROUTER_MODEL_ALIASES={"deepseek/deepseek-v4-flash":"deepseek-v4-flash"}
+
 # --------------------------------------------------------------------
 # Server.
 # --------------------------------------------------------------------
--- a/cmd/AGENTS.md
+++ b/cmd/AGENTS.md
@@ -15,6 +15,7 @@ Composition root. Only place that constructs concrete adapters + wires them toge
   - `runSessionPinSweep` — TTL sweep loop
   - `resolveHardPinModel` / `resolveCompactionModel` / `resolveDefaultBaselineModel` / `resolveAvailableModels` — boot-time model resolution
   - `registerDeploymentKeyedProvider` — shared "resolve key → build client → log" registration for the providers whose gating collapses to that shape (Fireworks, Makora, Together, Bedrock, Google); OpenRouter and Anthropic/OpenAI stay bespoke
+  - `resolveModelAliases` ([`model_aliases.go`](router/model_aliases.go)) — each OpenAI-compatible provider's outbound model ID map: catalog `UpstreamID` bindings with `ROUTER_<PROVIDER>_MODEL_ALIASES` overrides layered on top
   - small env parsers (e.g. `envVarHint`, `parseEnvInt`, `parseEnvFloat`, `parseEnvDurationMs`)
 - **No more heuristic-fallback router.** If cluster routing fails to boot, `main.go` panics. Misconfiguration must abort the process rather than silently degrade.
 - Validate deployment target overrides through `policy.Registry.ValidateDeployment` after provider/model configuration is resolved. A partial or catalog-incompatible explicit target is a startup error, not permission to substitute a default.
--- a/cmd/CLAUDE.md
+++ b/cmd/CLAUDE.md
@@ -15,6 +15,7 @@ Composition root. Only place that constructs concrete adapters + wires them toge
   - `runSessionPinSweep` — TTL sweep loop
   - `resolveHardPinModel` / `resolveCompactionModel` / `resolveDefaultBaselineModel` / `resolveAvailableModels` — boot-time model resolution
   - `registerDeploymentKeyedProvider` — shared "resolve key → build client → log" registration for the providers whose gating collapses to that shape (Fireworks, Makora, Together, Bedrock, Google); OpenRouter and Anthropic/OpenAI stay bespoke
+  - `resolveModelAliases` ([`model_aliases.go`](router/model_aliases.go)) — each OpenAI-compatible provider's outbound model ID map: catalog `UpstreamID` bindings with `ROUTER_<PROVIDER>_MODEL_ALIASES` overrides layered on top
   - small env parsers (e.g. `envVarHint`, `parseEnvInt`, `parseEnvFloat`, `parseEnvDurationMs`)
 - **No more heuristic-fallback router.** If cluster routing fails to boot, `main.go` panics. Misconfiguration must abort the process rather than silently degrade.
 - Validate deployment target overrides through `policy.Registry.ValidateDeployment` after provider/model configuration is resolved. A partial or catalog-incompatible explicit target is a startup error, not permission to substitute a default.
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -199,6 +199,11 @@ func main() {
 	// platform-key mode gated by balance checks. Self-hosted is never BYOK-only.
 	byokOnly := deploymentMode == server.DeploymentModeManaged && billingSvc == nil
 
+	modelAliases, err := resolveModelAliases(logger)
+	if err != nil {
+		panic(err)
+	}
+
 	// Always registered. With ANTHROPIC_API_KEY (selfhosted only) the router
 	// uses its own key; otherwise client auth headers pass through directly.
 	anthropicKey := ""
@@ -264,7 +269,7 @@ func main() {
 		if !byokOnly && openRouterPlatformEnabled {
 			openRouterKey = config.GetOr("OPENROUTER_API_KEY", "")
 		}
-		providerMap[providers.ProviderOpenRouter] = openaiCompatProvider.NewClient(openRouterKey, openRouterBaseURL)
+		providerMap[providers.ProviderOpenRouter] = openaiCompatProvider.NewClientWithModelIDMap(openRouterKey, openRouterBaseURL, modelAliases[providers.ProviderOpenRouter])
 		switch {
 		case byokOnly:
 			logger.Info("OpenRouter provider enabled (BYOK only)", "base_url", openRouterBaseURL)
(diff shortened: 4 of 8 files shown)


vignesh-chaturvedi:
Worth flagging prior art I should have led with: #557 already targets #526 and
#541 with the same core mechanism — an env var parsed at the composition root
and fed into `NewClientWithModelIDMap`. @Shangmin-Chen got there first.

The difference is scope. #557 adds `OPENROUTER_MODEL_ID_MAP`, which covers the
OpenCode Go case in #541. This PR generalizes to `ROUTER_<PROVIDER>_MODEL_ALIASES`
across all six OpenAI-compatible providers, which is what the second half of #526
asks for — Bedrock reached through an OpenAI-compatible proxy needs the same
remapping and can't use an OpenRouter-specific variable. It also layers over the
catalog's existing per-binding `UpstreamID` instead of sitting beside it, so
Fireworks/Together/Makora/Bedrock keep their built-in maps while an operator can
override a single entry.

If you'd rather land #557 and treat multi-provider support as a follow-up, I'm
happy to close this and rebase the generalization on top of it -- that ordering is
fine by me. Either way I wanted the overlap on the record rather than have two
PRs quietly competing.

vignesh-chaturvedi:
Closing this as superseded by #1381, which landed `ROUTER_MODEL_ID_MAP` and
closed #526. The OpenRouter case from #541 is covered there, so this PR no
longer has a reason to stay open.

One note in case it is useful later. `ROUTER_MODEL_ID_MAP` is passed to
`NewClientWithModelIDMap` on its own rather than layered over
`upstreamIDsForProvider(providers.ProviderOpenRouter)`, so the env map replaces
the catalog map instead of merging with it. That is inert today because
OpenRouter carries no per-binding `UpstreamID` entries, but if it ever gains
one, setting the env var would silently drop it. Merging the two maps with the
env entries taking precedence would keep that from becoming a surprise.

Thanks to everyone who looked at the issue.

PR #907, opened 2026-08-13
feat(catalog): add qwen/qwen3.8-max (Fireworks), retire qwen3.7-plus

aminsamir45:
## What
- Adds `qwen/qwen3.8-max` as a Fireworks-only TierHigh catalog entry (`accounts/fireworks/models/qwen3p8-max`, $2/$6 per MTok, $0.25 cache-read multiplier 0.125). Fireworks-only for the same reason `qwen3.7-plus` is: it's SOC-2, and OpenRouter/Together's routes for this model forward to Alibaba/DashScope.
- Retires `qwen/qwen3.7-plus` from routing (superseded), demoted to a no-tier priced passthrough — same pattern used for `claude-opus-4-8` — so lingering BYOK/direct pins still bill at real cost.
- Adds `/force-model` aliases: `qwen3.7-plus` keeps its own-name alias (passthrough only now); `qwen-max`/`qwen3.8-max`/`qwen3.8` point at the new model.
- Regenerated client-side pricing artifacts via `go run ./cmd/genprices` (`install/cc-statusline.sh`, `install/install.sh`, `install/pi-router/src/pricing.generated.ts`).
- Fixed `TestCandidateK12Loads`'s retired-model count (19 → 18 of 21) — this legacy cluster bundle predates the catalog demotion, matching the existing `deepseek-v4-pro`/`claude-opus-4-8` exclusions already carved out there.

## Why
Modal's own "Shared Endpoint" surface for this model (`modal.com/library/qwen/qwen3-8-max`) turned out to be a poor fit: no published/durable pricing, reads as a ~1-month promo, and would require standing up brand-new provider plumbing (per-workspace dynamic endpoint URL, proxy-token auth) for a single model. Fireworks already hosts the exact same model with published, stable pricing and zero new plumbing — it slots into the existing `openaicompat` Fireworks binding the same way `qwen3.7-plus` and `qwen3-coder` do.

## Testing
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Confirmed `qwen/qwen3.7-plus` was never in the HMM roster (only `qwen/qwen3-coder-next` is), so no roster removal was needed on that side.

## Stage C (HMM roster)
Handled separately in the WorkWeave monorepo (roster_v2/v4.json edit + neutral-prior bandit-state promotion + sidecar package republish), following the `hmm-add-model-prior` skill.


lowellh315:
Ship It!

PR #910, opened 2026-08-14
Show routing badge only when the served model changes

a-churchill:
## Summary

The routing badge fired on nearly every turn because the existing
suppression gate only triggered when the planner didn't run (tool-result
follow-ups). On a main turn the planner always reports a reason, so the
badge printed even when the model was identical to the prior turn —
20 of 22 badges in one observed session carried no information.

## Changes

- **internal/proxy/service.go**: rewrite the gate in `routingMarkerFor`
  to suppress whenever `PriorServedModel == decision.Model`, always
  emit on a genuine switch or on the first turn of a session/role.
  Suppress `HardPinned` turns explicitly (they never load the pin, so
  `PriorServedModel` is empty there). Sidecar `DisplayMarker` always
  prints — it's a genuine per-turn status line independent of whether
  the model changed. Apply the same same-model suppression to
  `baselineRoutingMarkerFor` so a failover landing back on the
  already-serving model stays quiet. `markerReasonHardPinned` becomes
  dead and is removed.
- **internal/proxy/marker_internal_test.go**: flip the four stay-reason
  cases (`ev_negative`, `no_prior_usage`, `same_tier_pinned`,
  `same_model`) to set a matching `PriorServedModel` and assert the
  marker is suppressed; invert the hard-pin case to assert suppression;
  add two new unit tests for `baselineRoutingMarkerFor`'s new gate.

### DIFF
--- a/internal/proxy/marker_internal_test.go
+++ b/internal/proxy/marker_internal_test.go
@@ -37,59 +37,43 @@ func TestRoutingMarkerFor_PlannerPaths(t *testing.T) {
 			},
 		},
 		{
-			name: "ev_negative: planner stayed",
+			// Stay reasons only fire when the scorer/planner re-confirmed the same
+			// model already serving this session — which is exactly the case the
+			// switch-only gate now suppresses.
+			name: "ev_negative: same model as last turn, marker suppressed",
 			res: turnLoopResult{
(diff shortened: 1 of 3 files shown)


a-churchill:
Tested locally

PR #918, opened 2026-08-15
Auto update Claude Code router config

a-churchill:
Makes two changes to the Claude Code router setup:
1. Uses your existing key if possible when running `npx @workweave/router` for an existing install
2. Auto updates the installed files (e.g. skill definitions)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ test-statusline: ## Run the cc-statusline.sh regression tests (offline)
 
 test-install: ## Run offline installer regression tests
 	@bash install/tests/codex_install_test.sh
+	@bash install/tests/key_reuse_test.sh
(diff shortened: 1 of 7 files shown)


a-churchill:
Fixed in 925f89c9: `update` now reads the base URL already configured (checking the parked sidecar first if toggled off) and keeps it unless `--base-url`/`--local` is passed explicitly, instead of resetting to the hosted default.

Also fixed the other three items from this review round in the same commit: parked-sidecar key reuse, the tracked-wrapper git-dirtying refresh, and the quiet-update nonzero-exit-on-rejection bug.

steventohme (dismissed):
LGTM. I think we gotta fix the devin comment tho

PR #949, opened 2026-08-19
fix: point star history chart at the dera.page mirror

OctoBored:
The star history chart is currently broken because the upstream star-history.com service is subject to GitHub stargazer API restrictions. This PR re-points the chart at the star-history.dera.page mirror, which serves the same data from a token-free source on a single domain, restoring a working star history chart for workweave/router.


steventohme:
THANK YOU GOAT @OctoBored

PR #950, opened 2026-08-19
feat(catalog): add routable deepseek-v4-pro-0813

aminsamir45:
feat(catalog): add routable deepseek-v4-pro-0813

The bare `deepseek/deepseek-v4-pro` ID resolves to the retired 0423 build and
was deliberately untiered in #896 ("Retire deepseek-v4-pro from routing; fail
over on upstream 402"). TierUnknown means "not deployable as a routing
target", so anything rostering that ID is inert -- it never appears in an
eligibility set and can never be served.

But the model Artificial Analysis actually scores is the 0813 release
(https://artificialanalysis.ai/models/deepseek-v4-pro -> "DeepSeek V4 Pro
0813", deprecated: false, agentic index 49.56). Our own aa_skill_scores entry
already records aa_name "DeepSeek V4 Pro 0813 (Reasoning, Max Effort)" with
that exact score, so the HMM roster has been ranking on 0813 while binding to
0423.

Adds `deepseek/deepseek-v4-pro-0813` as TierMid with its OpenRouter binding
($0.66/$1.98, cache read $0.022). The bare alias stays untiered so session
pins and /force-model keep dispatching to it, and the #896 retirement of the
0423 build is preserved.

Regenerated the pricing tables (cmd/genprices); TestStatuslinePricingMatchesTable
and TestCheckedInPiPricingIsCurrent both failed before this, which is the gate
that catches an unpriceable routing target.

Verified: DeployedModelsForRosterIDs resolves the 0813 ID with tier=mid; full
`go test ./...` green.

fix: address PR review feedback on deepseek-v4-pro-0813

Fixed:
- Trimmed the 5-line comment above the 0813 entry to 2 lines
  (workweave-bot comment-length nit)
- Added a Fireworks binding (accounts/fireworks/models/deepseek-v4-pro-0813,
  $1.32/$3.96, cache-read $0.044) ahead of the existing OpenRouter binding
  (cursor bugbot). ManagedProviderPolicy() denies OpenRouter for HMM/RL
  candidate resolution, and prod/staging both run
  ROUTER_DEFAULT_STRATEGY=hmm, so an OpenRouter-only binding was never
  reachable via managed routing -- the same 'never served' failure this PR
  set out to fix. Fireworks already hosts the model live; matches the
  sibling deepseek/deepseek-v4-pro entry's dual-binding pattern.
  Regenerated cc-statusline.sh/install.sh/pricing.generated.ts via
  cmd/genprices since the primary binding (and its price) changed.
- Bumped install/npm/package.json 0.2.10 -> 0.2.11: the already-published
  0.2.10 tarball predates this PR's pricing changes (greptile top-level
  comment), same pattern as PR #907's 0.2.7 -> 0.2.8 bump.

Declined (with explanation):
- greptile: statusline/Pi cache-read cost calculators hard-code a 0.1x
  multiplier instead of reading catalog.CacheReadMultiplier per model.
  Real but pre-existing and systemic -- roughly two dozen catalog entries
  already carry a non-0.1 CacheReadMultiplier, none of which are threaded
  through to the client calculators today. Fixing it means plumbing a
  per-model multiplier through cmd/genprices into both
  pricing.generated.ts and cc-statusline.sh's jq math, which is a
  standalone change, not something this single-model catalog addition
  should carry.


aminsamir45:
* **#950** <a href="https://app.graphite.com/github/pr/workweave/router/950?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/950?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #955, opened 2026-08-19
feat(routing): org positive model allowlist

aminsamir45:
## What

Adds an org-level **positive** model allowlist (`allowed_models` on `model_router_installations`), the counterpart to today's `excluded_models` deny list. The two compose: **effective set = allowlist − exclusions**. An empty allowlist means *no restriction*, not *no models*.

## How

**The allowlist is desugared into the exclusion set** at the proxy boundary rather than filtered separately. `excludedModelsForRequest` adds every routable model absent from a non-empty allowlist to the exclusion set, so all six existing enforcement sites — cluster scorer, policy resolver, session pins, sibling failover, force-model, fallback — honor it with no new filter loops.

`router.Request.AllowedModels` exists only so errors and diagnostics can *name* the allowlist instead of dumping a 50-entry desugared exclusion list.

## Two paths that needed explicit handling

Both intentionally bypass `excluded_models`, so neither would have honored the allowlist for free:

1. **`usageBypassEngaged`** consults `SafetyExcludedModels`, not `ExcludedModels`, on the documented rationale that exclusions are a routing preference the bypass may override. An allowlist is a compliance boundary, not a preference — without an explicit check the allowlist was **silently inert for all subscription-backed traffic**. `TestUsageBypass_NotAllowlistedModel_EngagesRouting` covers this and fails without the fix.

2. **`ROUTER_EXCLUDED_MODELS`** short-circuits `excludedModelsForRequest` entirely. Left as-is deliberately — an operator debugging a deployment shouldn't be constrained by one org's config — but now commented as an intentional gap rather than an accident.

## Error surfaces

- `ErrAllowlistEmptiesPool` wraps `ErrNoEligibleProvider`, so existing `errors.Is` checks keep matching while the message names the allowlist.
- New `not_allowlisted` resolver diagnostic, checked ahead of `requested_exclusion`.
- Dispatch-error class ordered **before** the generic no-eligible-provider case — `errors.Is` would otherwise match the wrapper first and tell an admin they have no provider keys.

## Admin API

`GET`/`PUT /admin/v1/allowed-models`, mirroring the excluded-models handlers.

## Validation

- `go build ./...`, `go vet ./...`, `go test ./...` all clean
- Migration applied and rolled back both directions locally
- The two behavioral tests were each verified to **fail** with the fix reverted

Part 1 of 2; per-user per-cluster selection follows in a stacked PR.

### DIFF
--- a/db/migrations/0057_installation-allowed-models.down.sql
+++ b/db/migrations/0057_installation-allowed-models.down.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE router.model_router_installations
+  DROP COLUMN allowed_models;
+
+COMMIT;
--- a/db/migrations/0057_installation-allowed-models.up.sql
+++ b/db/migrations/0057_installation-allowed-models.up.sql
@@ -1,0 +1,13 @@
+BEGIN;
+
+-- Positive counterpart to excluded_models: when non-empty, routing is confined
+-- to these models. Composes with (does not replace) excluded_models — the
+-- effective set is allowed_models minus excluded_models. Empty array means "no
+-- restriction", matching excluded_models' empty-means-none convention, so NULL
+-- and empty never have to be distinguished at the read site.
+ALTER TABLE router.model_router_installations
+  ADD COLUMN allowed_models TEXT[] NOT NULL DEFAULT '{}';
+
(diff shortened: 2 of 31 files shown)


aminsamir45:
* **#956** <a href="https://app.graphite.com/github/pr/workweave/router/956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#955** <a href="https://app.graphite.com/github/pr/workweave/router/955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #956, opened 2026-08-19
feat(routing): per-user per-cluster model selection

aminsamir45:
> Stacked on #955 — review that first.

## What

Lets an individual engineer choose which models they get **per HMM routing cluster** (`fast`/`explore`/`balanced`/`high`/`maximum`), within the bounds their org admin set.

## Why a new table

`cluster_model_lists` already implements per-cluster restriction and is fully enforced — but it is keyed on `api_key_id`, and a key is explicitly *"an installation-scoped secret, not a per-user credential"* (`0001_init.up.sql:77`) shared by every user on the installation. It cannot express a per-user selection.

`model_router_user_cluster_model_lists` is keyed on `router_user_id`. The two compose: **key-scoped = org default, user-scoped = personal narrowing.**

## Composition is intersection, not override

`mergeClusterOverrides` intersects the user's selection with the key-scoped list, preserving the user's ordering. This is the load-bearing decision: a plain "user wins" override would let an individual **re-admit a model the org deliberately removed** — privilege escalation through an admin control. An empty intersection (stale personal pick) falls back to the org list rather than emptying the cluster.

The org allowlist from #955 needs no handling here: it's desugared into the exclusion set upstream, and `effectiveArms` already intersects overrides against request-resolved eligible candidates.

## Where selections load

`ResolveAndStashUser`, not the auth middleware — user identity isn't resolved until the handler. Critically, on **both** the cache-hit and upsert paths: the identity cache short-circuits before the upsert, so loading only on the upsert path would silently drop every selection once the cache warmed. `TestResolveAndStashUser_StashesClusterListsOnIdentityCacheHit` fails if that path is reverted.

## Caching and invalidation

New `LRUUserClusterListCache` keyed by router user (per-user state can't live in the API-key cache), structurally mirroring `LRUAPIKeyCache` including its `byInstallation` index and `invalidationGen` race guard. `NewInvalidationListener` is now **variadic** so the existing `router-installation-invalidate` message fans out to both caches — without that wiring a user's own change would wait out the full TTL and read as "the setting doesn't work."

TTL is 5min, matched to the API-key cache rather than the 10min identity cache, so both halves of the feature share one staleness bound.

Fail-open on fetch error (serve default routing, don't cache); a successful **empty** result *is* cached, since most users configure nothing and re-querying them every request is the common case.

## Fail-open vs fail-closed

Deliberately asymmetric with the allowlist, and documented in `internal/proxy/CLAUDE.md` + `AGENTS.md`: an org allowlist is a compliance control (fail-closed), a personal selection is a preference (fail-open). This is only safe because the allowlist binds *upstream* in the resolver while the preference binds *downstream* in the override layer — so falling open there falls back to an already-allowlist-constrained set. **The allowlist must never be enforced in the override layer.**

## Capability flag

`HonorsClusterModelLists`, derived from `ReportsRankedFallback` (its precondition), so the control plane can hide the per-cluster UI for strategies that would silently ignore it — the legacy cluster scorer reports `false`.

## Validation

- `go build ./...`, `go vet ./...`, `go test ./...` all clean
- Migration applied and rolled back both directions locally
- Composition rules covered by `cluster_overrides_internal_test.go`, including the privilege-escalation guard


aminsamir45:
* **#956** <a href="https://app.graphite.com/github/pr/workweave/router/956?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/956?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#955** <a href="https://app.graphite.com/github/pr/workweave/router/955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #957, opened 2026-08-19
feat: add per-installation toggle to hide router terminal surfaces

steventohme:
## What

Adds an organization-specific boolean, `hide_terminal_surfaces` (default `false`), on `router.model_router_installations`. When an org turns it on, the router suppresses the surfaces it injects into engineers' terminals while **routing and billing are unchanged**:

- the `✦ Weave Router → <model> · <reason>` routing marker
- the trailing `_Weave Router feedback:_ \`/rf +\` … \`/rf -\`` hint
- the `WEAVE ROUTER — <routed> ← <selected> · saved $X` statusline

Hiding only suppresses the **injected prompts** — explicit `/rf` use and signed rating links still accept and record feedback (the no-login feedback API and DB persistence stay live). The setting is **silent**: engineers see a normal terminal, no indication that requests are routed.

The flag is stored on the installation row and written by the platform (WorkWeave) via the shared DB — the same pattern as `subscription_routing_disabled`. A companion WorkWeave PR adds the GraphQL mutation + settings UI.

## How

- **Migration `0055`** + `UpdateModelRouterInstallationHideTerminalSurfaces` sqlc query; `internal/sqlc/` regenerated via `make generate`.
- **Domain/adapter/service**: `auth.Installation.HideTerminalSurfaces`, `InstallationRepository.UpdateHideTerminalSurfaces`, postgres adapter + converter, and `auth.Service.SetInstallationHideTerminalSurfaces` (invalidates the auth cache).
- **Request path**: auth middleware stashes `proxy.InstallationHideTerminalSurfacesContextKey`; `hideTerminalSurfacesForRequest` reads it. Gated at the three choke points — `suppressMarkerIfRequested` (covers every routing-badge call site), `feedbackFooter`, and `setFeedbackLinkHeader`. The subscription-only depleted-credits warning marker is deliberately untouched (billing CTA, not routing info).
- **Statusline/installer**: new bearer-authenticated `GET /v1/display-settings` returns `{ hide_terminal_surfaces }` for the calling installation. `cc-statusline.sh` (and its byte-identical copy embedded in `install.sh`) checks it — cached per install under `~/.cache/weave-router` with a TTL, fail-open on any error — and prints nothing when the org hides surfaces. Existing installs pick this up via the script's weekly self-refresh; new installs check at install time.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` — all green.
- `make generate` — no drift.
- Migration `0055` verified up → down → re-up against a local router DB.
- `install/tests/cc-statusline_test.sh` — 30/30 pass (including the byte-identical heredoc check), plus a new hidden-suppression case for the footer and marker.

## Deploy order

Router first (migration + reads + gating), then bump the WorkWeave gitlink, then the platform PR (GraphQL + settings UI). The column defaults to `FALSE`, so each side tolerates the other not being deployed yet.

### DIFF
--- a/db/migrations/0055_installation-hide-terminal-surfaces.down.sql
+++ b/db/migrations/0055_installation-hide-terminal-surfaces.down.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE router.model_router_installations
+  DROP COLUMN hide_terminal_surfaces;
+
+COMMIT;
--- a/db/migrations/0055_installation-hide-terminal-surfaces.up.sql
+++ b/db/migrations/0055_installation-hide-terminal-surfaces.up.sql
@@ -1,0 +1,13 @@
+BEGIN;
+
+-- Per-installation toggle to hide the router's terminal surfaces from
+-- engineers: the "✦ Weave Router → <model>" routing marker, the trailing
+-- "/rf" feedback hint, and the router statusline. When true, requests route
+-- identically but nothing routing-related is rendered in the caller's
+-- terminal. Feedback submission/recording (/rf, signed rating links) is
+-- unaffected. Default false preserves today's behavior for every existing
+-- installation.
+ALTER TABLE router.model_router_installations
+  ADD COLUMN hide_terminal_surfaces BOOLEAN NOT NULL DEFAULT FALSE;
+
+COMMIT;
--- a/db/queries/model_router_installations.sql
+++ b/db/queries/model_router_installations.sql
@@ -102,3 +102,15 @@ SET subscription_routing_disabled = @subscription_routing_disabled::boolean,
 WHERE id = @id::uuid
   AND external_id = @external_id::varchar
   AND deleted_at IS NULL;
+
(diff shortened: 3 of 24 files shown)


steventohme:
Addressed all review feedback in `b20dbeda`:

**Statusline gate correctness** (cursor high/medium, greptile P1):
- `weave_hidden_gate` now runs with stdin detached (`</dev/null` at the call site), so a cache-miss curl can't drain the statusline JSON Claude Code pipes in.
- Credentials for `/v1/display-settings` resolve relative to the script's own location (`<base>/.claude` for project/--dir installs) instead of always `~/.claude`, so a project install uses its own key and never leaks the user-scope key. `file://` base URLs read as the body directly (offline/test seam); https URLs get `/v1/display-settings` appended.
- Added regression tests: project-scope hidden org renders blank, project install doesn't fall back to the user key, visible org still renders. 33/33 statusline tests pass.

**Comment-length nits** (workweave-bot): applied all 8 suggestions verbatim.

PR #965, opened 2026-08-20
fix(translate): hoist tool_result images into the OpenAI user message

aminsamir45:
Images nested inside an Anthropic tool_result block (e.g. a screenshot
returned by a Read tool call) were silently dropped on the OpenAI-compat
emit path: role:tool content is text-only, so buildOpenAIToolResultMessage
flattened them to an empty string and the upstream model never saw the
image — sessions then burned turns on OCR workarounds for content they
already had.

writeOpenAIUserFromAnthropic already splits mixed tool_result + text +
image content into tool messages followed by a user message; it now also
hoists nested tool_result images into that trailing user message as
image_url parts (the same shape top-level user images already use).

Detection was never broken: anthropicImageBytes counts tool_result-nested
images for HasImages/token math, so capability gates still routed
image-bearing turns to multimodal models — only the emit side lost them.

### DIFF
--- a/internal/translate/crossformat_request_test.go
+++ b/internal/translate/crossformat_request_test.go
@@ -921,10 +921,8 @@ func TestCrossFormat_AnthropicToOpenAI_Image(t *testing.T) {
 	assert.Contains(t, url, "iVBORw0KGgo=")
 }
 
-// Agent harnesses return screenshots nested inside tool_result blocks rather
-// than at the top level of the user message. OpenAI role:tool content is
-// text-only, so the image must be hoisted into the trailing user message —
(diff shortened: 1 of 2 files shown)


aminsamir45:
* **#965** <a href="https://app.graphite.com/github/pr/workweave/router/965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aminsamir45:
Applied both comment-trim suggestions verbatim in aa27cc79.

PR #977, opened 2026-08-20
fix(proxy): resolve spoken model slugs and list pinnable models on /force-model

aminsamir45:
Two /force-model failures made pinning unreliable, both silent:

1. `/fm qwen 3.8` pinned qwen/qwen3-coder. The parser split the argument
   at the first space, so only "qwen" reached the resolver — a valid alias
   for a different model — and "3.8" was dropped into the prompt. The ack
   read "force-model applied: qwen/qwen3-coder", so a user asking for one
   model was served another under a message that looked like success.
   `/fm qwen max` failed the same way.

2. There was no way to discover what could be pinned. A bare /force-model
   matched nothing and was forwarded upstream, so a model answered the
   question by guessing, and the unrecognized-model reply named three
   hardcoded examples that may not even be available on the installation.

The parser now resolves the model-name/prompt boundary through the same
resolver that performs the pin, taking the longest leading run of words
that resolves, so parser and pin cannot disagree about where the name
ends. When no multi-word run resolves, the name still absorbs following
version-like tokens: without that, "/fm qwen 9.9" falls back to the bare
alias "qwen" and silently pins the wrong model instead of rejecting a
model the user never named.

Resolution is also separator-insensitive, so "qwen 3.8", "qwen-3.8", and
"qwen3.8" all reach qwen/qwen3.8-max. This runs only after the exact
lookups, never overriding a precise match, and refuses a leading/trailing
separator so a truncated "gpt-" stays unknown rather than folding into
the "gpt" alias. Ambiguous folds are refused rather than guessed.

A bare /force-model now lists every pinnable model, split into routing
targets and passthrough-only rows, with each one's provider and
shorthands. The listing is derived from forcedModelBinding — the same
gate that admits a pin — so it cannot advertise a model that would be
refused a moment later. It deliberately is not the policy sidecar's
roster: models that are pinnable but never auto-selected, and models the
roster omits, are reachable and so are listed. Unrecognized models now
get the closest matching ids instead of fixed examples.

The listing ack is classified as a no-op in both the statusline and the
pi extension: it changes no pin state, and treating it as a clear would
drop a live pin from the UI. The "isn't a recognized model" sentinel the
statusline greps for is preserved verbatim.

Co-Authored-By: Weave Router <router@workweave.ai>

## Testing

- `gofmt -l .`, `go vet ./...`, `go build -o /dev/null ./...`, `go test -count=1 ./...` — all clean.
- `make test-install` — 3/3 suites pass (Codex, key reuse, models).
- `make test-statusline` — 31 pass, 2 fail; both reproduce on a pristine
  checkout of the base commit and concern refresh-stamp counts, untouched here.
- New coverage: multi-word and separator-insensitive resolution, the
  truncated-`gpt-` guard, listing membership/ordering/exclusion filtering,
  suggestion ranking, and end-to-end handler tests asserting on what the pin
  store actually recorded (the original bug was a wrong pin under a
  success-looking ack).
- Verified the statusline `jq` classifier and the pi extension's regexes
  against the real message text for all six ack shapes.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -343,13 +343,28 @@ with nowhere to go (HTTP 503 from the scorer), so exclude deliberately.
 
 ## Forcing a model or a routing cluster
 
+`/force-model <model>` (alias `/fm`) pins the session to one model. The model
+may be a canonical catalog ID, an alias (`opus`, `qwen-max`, …), or the way you
+would say it out loud — separators are insignificant, so `qwen 3.8`,
+`qwen-3.8`, and `qwen3.8` all reach `qwen/qwen3.8-max`. Anything after the
+model name stays in the turn as your prompt (`/fm qwen 3.8 fix the test`).
+
+Run `/force-model` with no argument to list every model this installation can
+pin, grouped into automatic-routing targets and passthrough-only models, with
+each one's provider and shorthands. The listing is derived from the same gate
+that admits a pin, so anything it names will be accepted. It is not the policy
+sidecar's roster: models that are pinnable but never chosen automatically —
+and models the sidecar's roster omits — appear too, because you can reach
+them. An unrecognized model is refused with the closest matching IDs rather
+than pinned, and any prior pin is left alone.
+
 Two request headers let a headless caller (eval harness, CI, any client whose
 UI eats slash commands) override routing. Both fail the request rather than
 routing on, so a typo can't look like it took effect.
 
 | Header | Effect |
 | ------ | ------ |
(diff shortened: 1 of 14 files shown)


aminsamir45:
* **#977** <a href="https://app.graphite.com/github/pr/workweave/router/977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #985, opened 2026-08-21
fix(proxy): route grok/xai pins to 4.6 and floor grok effort at low

aminsamir45:
Force-model fixes for the grok family, prompted by a bad `/fm grok` pin that silently resolved to the retired 4.5 and then ran unusably slow.

## The problem

`/fm grok` pinned **grok-4.5**. Two things made that a bad outcome:

1. **Stale family alias.** `grok` / `xai` resolved to grok-4.5, which is retired from routing (no AA Agentic Index score, never rostered) and underperformed as a pin. Family aliases should follow the current flagship, the way `gpt`/`gemini` do.
2. **A bare grok pin sends no `reasoning_effort` at all**, and xAI's server-side default for grok-4.x is **high and not disableable**. Measured 2026-08-21 (prod, claude-code, input >20k): grok-4.6 p50 TTFT **15,444 ms** vs 2,949 ms for the next-worst arm — a tight p50/p90 band (15,444/15,985) marking a fixed reasoning stall, not variable prefill, and only ~107 avg output tokens.

## The fix

- **`grok` / `xai` → grok-4.6.** grok-4.5 stays servable as priced passthrough via its own-name alias, so any existing explicit `grok-4.5` pin keeps resolving.
- **Pin grok-4.x to `low` in `forcedReasoningEffort`, unconditionally.** This is the dispatch-time resolver hit on every forced and sticky-pin turn (forced, baseline, sibling seams), so a bare pin — header **or** chat command — no longer falls to xAI's high default. An explicit `:level` suffix still wins (it lands on `ForceEffort`, which outranks `ForceReasoningEffort`).

Two deliberate placement choices, both correcting earlier drafts:

- **Dispatch-time, not resolve-time.** Force-model resolution (the header's `resolveForceModelWithEffort` and the chat command's `resolveForceModel`) runs only on the turn that sets the pin; sticky-pin turns never re-derive effort. A resolve-time floor would miss the exact pinned turns this is for. `forcedReasoningEffort` runs at every dispatch seam.
- **Unconditional, not behind `ROUTER_EFFORT_ESCALATION`.** That flag gates the gpt-5/gemini-3 escalate-on-failure *policy* and defaults off. The grok floor is a correction for a defect (a bare pin falling to a non-disableable high default), not an escalation to bake off, so it applies regardless of the flag. The escalation policy stays gated.

## Behavior

```
/fm grok          -> grok-4.6, effort=low   (was: grok-4.5, no effort -> xAI high)
/fm xai           -> grok-4.6, effort=low   (was: grok-4.5)
/fm grok-4.5      -> grok-4.5, effort=low   (passthrough, still resolvable)
/fm grok-4.6:high -> grok-4.6, effort=high  (explicit suffix wins)
sticky grok pin   -> effort=low every turn  (was: no effort -> xAI high), flag on or off
```

## Testing

- `gofmt -l .` clean, `go vet ./...`, `go build ./...`, `go test ./...` — all pass.
- Coverage: `TestResolveForceModel_GrokFamilyAlias` (grok/xai → 4.6), `TestResolveForceModel_EffortSuffixPreserved` (`:high` survives), and new `forcedReasoningEffort` cases asserting grok-4.5/4.6 pin low with the floor ignoring escalation.

### DIFF
--- a/internal/proxy/force_effort_test.go
+++ b/internal/proxy/force_effort_test.go
@@ -17,8 +17,8 @@ func TestForcedReasoningEffort(t *testing.T) {
 		{"gpt-5.4-mini", true, "high"},
 		{"gemini-3.1-pro-preview", false, "low"},
 		{"gemini-3.1-pro-preview", true, "low"}, // effort-immune: escalation ignored
-		{"grok-4.6", false, "low"},              // bare pin must not fall to xAI's non-disableable high default
-		{"grok-4.6", true, "low"},               // pinned low; escalation not spent on grok
+		{"grok-4.6", false, "low"},              // unconditional: bare pin must not fall to xAI's non-disableable high default
+		{"grok-4.6", true, "low"},               // floor ignores escalation
 		{"grok-4.5", false, "low"},
 		{"claude-opus-4-8", false, ""}, // adaptive path untouched
 		{"claude-opus-4-8", true, ""},
--- a/internal/proxy/force_effort_test.go
+++ b/internal/proxy/force_effort_test.go
@@ -17,7 +17,10 @@ func TestForcedReasoningEffort(t *testing.T) {
 		{"gpt-5.4-mini", true, "high"},
 		{"gemini-3.1-pro-preview", false, "low"},
 		{"gemini-3.1-pro-preview", true, "low"}, // effort-immune: escalation ignored
-		{"claude-opus-4-8", false, ""},          // adaptive path untouched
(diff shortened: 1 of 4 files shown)


aminsamir45:
* **#985** <a href="https://app.graphite.com/github/pr/workweave/router/985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/985?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #987, opened 2026-08-21
fix(translate,proxy): intersect allOf branches; never hard-fail a turn on an untranslatable tool schema

aminsamir45:
## The bug

Every request carrying Claude Code's `SendMessage` tool that the policy routed to Gemini failed **before dispatch**:

```
translate anthropic request to gemini: function "SendMessage" input_schema:
gemini schema is not representable at $.properties.to.allOf[1]: branches conflict
```

Prod, org Weave (`org_QWsHDcRQWQEs6RpkdEZrlFK8`, key `weave-checks`):

| Day | Failures |
|---|---|
| 2026-08-19 | ≥5,000 (query cap) |
| 2026-08-20 | **9,337** |
| 2026-08-21 (to 18:20Z) | 3,915 |

761 of the 849 affected sessions failed **exactly 11 times** — the client's full retry ladder against a request no upstream ever saw. So ~875 real user turns died, each burning 11 round trips.

### 1. `allOf` was treated as "must be identical", not "must both hold"

`mergeSchemaMapsExact` merged `properties` and `required`, then required every other shared keyword to be byte-equal. `allOf` means *intersection*, so branches differing on `pattern`, `description`, or a bound are the normal case, not a conflict.

`intersectSchemaKeyword` now intersects per keyword:

| Keyword | Intersection |
|---|---|
| `minimum`/`minLength`/`minItems`/`minProperties` | larger (stricter) |
| `maximum`/`maxLength`/`maxItems`/`maxProperties` | smaller (stricter) |
| `enum` | set intersection; empty → reject |
| `type` | equal → keep; `integer`∧`number` → `integer`; else reject |
| `items` | recurse as a schema |
| `nullable` | AND, reconciled from both inputs |
| `description`/`title`/`default`/`example`/`propertyOrdering` | annotations — take one side |
| `pattern`/`format` | Gemini carries one; keep the left |
| anything else | unchanged equality rule |

Only a genuinely empty intersection still rejects (disjoint enums, two different types) — the existing `{"type":"string"}` ∧ `{"type":"number"}` rejection test is untouched.

`pattern`/`format` is the one lossy case: keeping one **widens** the accepted input language. That is the same deliberate trade already made for `exclusiveMinimum → minimum` and for dropping unsupported `format` values, and it is commented as such. The tool's own handler still validates its arguments.

### 2. `properties` was merged as though its keys were schema keywords

`properties` is a map of property *names* → schemas, but it was fed straight back into `mergeSchemaMapsExact`, so two branches constraining the same property differently were compared for **equality** and rejected. Nested merging only ever worked for disjoint property sets. `intersectSchemaProperties` now recurses per property name.

### 3. A translation failure hard-failed the turn, invisibly

`buildAttempt`'s error returned straight out of `ProxyMessages` — before the baseline/sibling failover eligibility block, and before `fireTelemetry`. Two consequences:

- The turn died even though **the same tools translate fine for Anthropic**. A schema the routed provider cannot express is a property of the *route*, not of the request.
- No telemetry row at all. The affected key showed **5,290 `decision_provider='google'` rows with 0 errors** while ~70% of its google-routed requests were failing. The dashboard read 100% healthy.

Now: the turn is rescued on the baseline pre-dispatch (nothing has reached the wire, so the prelude is still discardable), the failure is recorded in telemetry with the status the client will actually receive, and with no baseline to rescue to it surfaces as a classified **400** instead of the unclassified **502 "Upstream call failed."** — which is what invited the eleven retries for a request that never left the router.

## Tests

- 9 new `TestPrepareGemini_SchemaFidelity` cases, including the incident's shape (two `pattern`s under `allOf`), disjoint enums still rejecting, `integer`∧`number`, ANDed nullability, and recursive nested-property merging.
- `TestProxyMessages_UntranslatableToolSchemaRescuesToBaseline` — asserts the Gemini upstream is **never** called and the turn serves on Anthropic with the caller's model.
- `TestProxyMessages_UntranslatableToolSchemaWithNoBaselineIsClassified` — no baseline still fails, but as a classified 400 with no `Retry-After`.
- `TestClassifyDispatchError_UntranslatableToolSchema` — pins 400 / client-error / no-retry / names the tool.

## Validation

`wv mr tc`, `wv mr t` (full suite), `go vet ./...` — all clean.

## Follow-up (not in this PR)

Make untranslatable-tools a *routing* signal so the policy excludes Gemini for a request carrying such a tool, instead of picking it and then rescuing. Avoids a wasted decision and the pin churn.

### DIFF
--- a/internal/proxy/dispatch_error.go
+++ b/internal/proxy/dispatch_error.go
@@ -50,7 +50,17 @@ const (
 	DispatchErrorForcedModelUnknown
 	DispatchErrorForcedClusterUnsupportedStrategy
 	DispatchErrorForcedClusterUnservable
+	DispatchErrorToolSchemaUntranslatable
 )
+
+// isUntranslatableToolSchemaErr reports whether err is a tool schema the routed
+// provider's function-calling grammar cannot express. Pre-dispatch and
+// route-specific: the same tools translate for another provider, so the caller
+// can rescue the turn instead of failing it.
+func isUntranslatableToolSchemaErr(err error) bool {
+	return errors.Is(err, translate.ErrGeminiSchemaIncompatible) ||
+		errors.Is(err, translate.ErrGeminiToolDeclarationConflict)
+}
 
 // DispatchErrorClass is the format-agnostic classification of a dispatch
 // error: the HTTP status to return, the client-facing message, whether to
@@ -161,6 +171,18 @@ func ClassifyDispatchError(err error) (DispatchErrorClass, bool) {
 			Message:    "This request requires a native compatibility path that the router does not support.",
 			LogLevel:   "warn",
 			LogMessage: "Translation requirements are intrinsically incompatible",
+		}, true
+	case isUntranslatableToolSchemaErr(err):
+		// 400, not 502: nothing was dispatched and retrying is futile. The
+		// unclassified 502 this used to fall through to told Claude Code the
+		// upstream had failed, so it retried the same doomed request eleven
+		// times per turn.
+		return DispatchErrorClass{
+			Kind:       DispatchErrorToolSchemaUntranslatable,
+			Status:     http.StatusBadRequest,
+			Message:    "A tool's input schema cannot be expressed in the routed provider's function-calling schema: " + err.Error() + ". Simplify that tool's schema, or exclude this provider for the installation.",
+			LogLevel:   "error",
+			LogMessage: "Rejected request: a tool schema is not representable for the routed provider",
 		}, true
 	case errors.Is(err, ErrTranslationCompatibleProviderUnavailable):
 		return DispatchErrorClass{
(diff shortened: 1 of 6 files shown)


aminsamir45:
Closing as duplicate. #986 (merged) already strips SendMessage from cross-vendor emit, which removes the prod symptom, and #980 covers the general allOf intersection plus the baseline failover with a more developed per-tool backstop. Deferring to those rather than landing a third overlapping change.

Two notes for #980's author, from the prod investigation behind this branch:

1. **Telemetry blind spot is not covered by any of the three PRs.** `buildAttempt`'s error returns from `ProxyMessages` before `fireTelemetry`, so a pre-dispatch translation failure writes no telemetry row. The affected key showed 5,290 `decision_provider='google'` rows with **0 errors** while ~9,337 requests/day were failing. Worth folding into #980 so the next translation bug is visible.

2. **`properties` merges as if its keys were schema keywords.** #980's `mergeSchemaMapsLenient` drops a conflicting branch wholesale — its own test asserts property `b` disappears when the branches conflict on a sibling `pattern`. Intersecting per property *name* (union of names, recursive merge on shared ones) preserves the tool's input language instead. Reasonable to treat as a deliberate trade, just worth being explicit about.

Data, if useful: 9,337 failures on 2026-08-20 for org Weave / key `weave-checks`; 761 of 849 sessions failed exactly 11 times each (the client retry ladder against an unclassified 502 for a request no upstream saw).

PR #1007, opened 2026-08-24
Add Wafer provider

a-churchill:
Tested locally


a-churchill:
Fixed in d0268e5c: Wafer-ZDR is now configured as a protected header and reapplied after prepared headers on both OpenAI-compatible and Anthropic routed/passthrough paths. Regression tests cover all four flows.

PR #1036, opened 2026-08-25
fix: persist OpenAI GPT-5.6 cache_write_tokens

aminsamir45:
GPT-5.6 Responses reports cache writes as `input_tokens_details.cache_write_tokens` (billed at 1.25×). We only parsed `cached_tokens` and hardcoded write=0, so telemetry, cost, and the Claude Code statusline showed **0 cache write** on every GPT turn even when cache was working (~97% read hit on the sessions that prompted this).

## What changed
- Shared `OpenAICacheTokens` helper reads `cache_write_tokens` (preferred) or `cache_creation_tokens` (compat fallback) from both Responses `input_tokens_details` and Chat Completions `prompt_tokens_details`.
- Responses → Anthropic streaming/non-streaming now forwards writes as `cache_creation_input_tokens` and subtracts them from fresh `input_tokens`.
- OTel usage extractor persists writes so `EffectiveInputCost` can apply the 1.25× multiplier instead of treating those tokens as 1.0× input.

## Context
Munir sessions `fd49cc93-…` / `85e6e58b-…` — GPT cache reads were fine; writes were dropped. Anthropic already reported both fields.

### DIFF
--- a/internal/observability/otel/usage.go
+++ b/internal/observability/otel/usage.go
@@ -105,8 +105,8 @@ func (u *UsageExtractor) RecordUsage(inputTokens, outputTokens int) {
 	}
 }
 
-// RecordCacheUsage sets cache token counts directly. OpenAI has no cache-creation
-// concept, so callers pass 0 for cacheCreationTokens.
+// RecordCacheUsage sets cache token counts directly. Pass 0 for a field the
+// provider does not emit.
 func (u *UsageExtractor) RecordCacheUsage(cacheCreationTokens, cacheReadTokens int) {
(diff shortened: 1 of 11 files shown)


aminsamir45:
* **#1036** <a href="https://app.graphite.com/github/pr/workweave/router/1036?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/1036?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1045, opened 2026-08-25
fix(admin): reject allowlists that empty the routed pool

aminsamir45:
## Summary
- Greptile was right: after #1041, `PUT /admin/v1/allowed-models` accepts any catalog ID. A non-empty allowlist is desugared as “exclude every model in `routableUniverse()` that isn’t listed,” so a list with no routable member (e.g. only `claude-opus-4-8` on a box that serves `claude-opus-4-7`) empties the candidate pool and every routed request 400s with `ErrAllowlistEmptiesPool`.
- Membership stays catalog-wide so force-model / hard-pin can still name non-roster rows. The new guard requires at least one ID in `proxy.Service.RoutableModels()` (the same set the desugaring uses). Empty list still clears the restriction. Unknown IDs still surface as `ErrUnknownModel` first.

## Test plan
- [x] `go test ./internal/api/admin/` (guard + HTTP 400)
- [x] `go test ./internal/proxy/` (wholly-unroutable allowlist empties the pool; `RoutableModels` matches the desugaring universe)
- [x] `wv mr tc` / `wv mr t`

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -945,7 +945,7 @@ func main() {
 		WithSummarizer(summarizer).
 		WithWebSearchExecutor(cortexWebSearch(logger)).
 		WithCompaction(compactionSz, compactionPct).
-		WithAvailableModels(routingTargets).
+		WithAvailableModels(proxyRoutableModels(routingTargets, availableProviders, hmmRouter != nil)).
 		WithDefaultBaselineModel(resolveDefaultBaselineModel()).
 		WithBillingService(billingSvc)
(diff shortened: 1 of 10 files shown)


aminsamir45:
* **#1045** <a href="https://app.graphite.com/github/pr/workweave/router/1045?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/1045?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1051, opened 2026-08-26
fix(openaicompat): surface the real gateway error after a version probe

aminsamir45:
## Summary

#1050 already drops leftover inbound `reasoning_effort` on OpenAI-format chat/completions emit. That is necessary but not sufficient for serving `gpt-5.6-luna` on an unversioned Cortex `openai_gateway` base (`…/api/v2/cortex`).

Every Cortex request first probes `{base}/chat/completions` (404, empty body), then hits `{base}/v1/chat/completions`. On a real 400 the adapter returned the **probe** 404, so Claude Code reported:

> There's an issue with the selected model (claude-opus-5[1m]). It may not exist or you may not have access to it.

while the useful Cortex body (`Function tools with reasoning_effort are not supported` / `unknown model "openai-1p-gpt-5.6-luna"`) never left the router.

This PR:

- returns the versioned-path error whenever it is not a 404
- memoizes that path so the next call skips the wasted probe

Anthropic-ingress emit already omits `reasoning_effort` for catalog `gpt-5*` (`TargetModel` stays `gpt-5.6-luna`; the Cortex alias is applied after emit). Combined with #1050, a luna turn on Cortex should now either succeed (no effort field) or fail with Cortex's real 400 instead of a fake model-not-found.

## Test plan

- [x] `go test ./internal/providers/openaicompat ./internal/providers`
- [ ] After merge + router pin bump: `sf ai claude -p "hi"` on the Cortex-backed key should not 404 as a missing `claude-opus-5[1m]`

### DIFF
--- a/internal/providers/openaicompat/client.go
+++ b/internal/providers/openaicompat/client.go
@@ -235,15 +235,17 @@ func (c *Client) Proxy(ctx context.Context, decision router.Decision, prep provi
 	baseURL := proxy.EffectiveBaseURL(ctx, c.baseURL)
 
 	// 404s are buffered before reaching w, so a missing "/v1" segment can be retried.
-	// firstErr is returned on double-miss so the probe's 404 doesn't mask the original.
+	// A non-404 on the versioned path is the real error and is memoized; only a
+	// second 404 falls back to the probe so a genuine model-not-found is preserved.
 	urls := c.versionMemo.URLs(baseURL, "/chat/completions")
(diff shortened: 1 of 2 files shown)


aminsamir45:
* **#1051** <a href="https://app.graphite.com/github/pr/workweave/router/1051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/1051?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1073, opened 2026-08-27
translate: find first user message regardless of leading system/developer prompt

Adityakk9031:
### Summary
Fixes `FirstUserMessageText` in `internal/translate/envelope.go` to find the first message with `role == "user"` by walking the messages array instead of only checking `messages.0`.

### Problem
In OpenAI Chat Completions and envelopes with leading `system` or `developer` prompts, `messages.0` is not `role == "user"`. As a result, `FirstUserMessageText()` returned an empty string `""`. When deriving session keys (`DeriveSessionKey`), this caused all OpenAI conversations under the same API key sharing a standard system prompt to derive identical session keys, leading to session pin collisions, cross-session pin thrashing, and degraded cache affinity.

### Changes
- Updated `FirstUserMessageText` to iterate through `messages` to find the first `role == "user"` block across all supported wire formats.
- Added test coverage in `internal/translate/envelope_extras_test.go` for leading system and developer messages.
- Added test in `internal/proxy/session_key_test.go` asserting that independent OpenAI conversations sharing a system prompt produce distinct session keys.
- Normalized CRLF line endings in `internal/proxy/conformance_golden_test.go` for cross-platform test stability.

### DIFF
--- a/internal/proxy/conformance_golden_test.go
+++ b/internal/proxy/conformance_golden_test.go
@@ -52,7 +52,9 @@ func golden(t *testing.T, name string, got []byte) {
 	}
 	want, err := os.ReadFile(path)
 	require.NoError(t, err, "missing golden %s — run `go test -update` to generate it", path)
-	require.Equal(t, string(want), string(got),
+	wantStr := strings.ReplaceAll(string(want), "\r\n", "\n")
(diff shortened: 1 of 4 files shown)


Adityakk9031:
@aminsamir45 have a look

PR #1074, opened 2026-08-27
middleware: exempt subscription turns from spend caps without requiring UsageBypassEnabled

Adityakk9031:
### Summary
Aligns `WithOrgMonthlySpendCap` and `WithAPIKeySpendCap` with `WithBalanceCheck` so that subscription-covered requests (which debit $0) are exempt from spend limits without requiring `UsageBypassEnabled`.

### Problem
In `WithBalanceCheck`, any request presenting a valid covering credential (e.g. `X-Weave-Anthropic-Subscription` or `X-Weave-OpenAI-Subscription`) passes through flagged `subscription-only` because subscription turns debit $0 on the caller's own plan.
However, `WithOrgMonthlySpendCap` and `WithAPIKeySpendCap` erroneously required `installation.UsageBypassEnabled && ...`. This caused organizations or keys with valid subscription credentials (but without the usage-bypass toggle active) to be rejected with HTTP 402 once their monthly or key limit was reached, even though the turn is free and incurs no platform cost.

### Changes
- Updated `WithOrgMonthlySpendCap` and `WithAPIKeySpendCap` to check `proxy.RequestPresentsCoveringSubscription(...)` directly, mirroring `WithBalanceCheck`.
- Updated unit tests in `org_monthly_spend_cap_test.go` and `api_key_spend_cap_test.go` to assert that covered requests pass through as subscription-only regardless of whether `UsageBypassEnabled` is enabled.

### DIFF
--- a/internal/server/middleware/api_key_spend_cap.go
+++ b/internal/server/middleware/api_key_spend_cap.go
@@ -38,14 +38,11 @@ func WithAPIKeySpendCap(svc *billing.Service) gin.HandlerFunc {
 			return
 		}
 
-		// The cap bounds PAID spend, not free subscription usage: a usage-bypass
-		// org presenting a Claude/Codex credential that covers this route serves at
+		// The cap bounds PAID spend, not free subscription usage: a request
+		// presenting a Claude/Codex credential that covers this route serves at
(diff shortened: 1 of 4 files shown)


Adityakk9031:
@aminsamir45 have a look

PR #1075, opened 2026-08-27
translate/gemini: fix tool call fragmentation across multi-chunk streaming deltas

Adityakk9031:
### Summary
Fixes function call streaming translation in `GeminiToOpenAISSETranslator` (`internal/translate/gemini_stream.go`) so multi-chunk streaming tool calls retain stable tool call IDs and indices without fragmenting into distinct duplicate tool calls.

### Problem
In `GeminiToOpenAISSETranslator.translateEvent`, receiving a `functionCall` part unconditionally minted a new tool call ID and incremented `toolIdx++`. When upstream Gemini models stream function call arguments across multiple SSE chunks (or send args continuation deltas without re-sending the function name), this caused a single function call to be emitted as multiple distinct OpenAI tool calls with incrementing indices and mismatched IDs, causing downstream client parser and execution failures.

### Changes
- Added active tool call tracking (`currentToolID`, `currentToolName`) to `GeminiToOpenAISSETranslator`.
- Tool index and IDs are preserved across partial argument deltas of the same function call.
- Added `emitToolCallArgsOnlyChunk` for args-only deltas.
- Added unit test `TestGeminiToOpenAISSETranslator_MultiChunkToolCall` asserting that multi-chunk function call streaming preserves index 0 and stable function attributes.

### DIFF
--- a/internal/translate/gemini_stream.go
+++ b/internal/translate/gemini_stream.go
@@ -35,7 +35,9 @@ type GeminiToOpenAISSETranslator struct {
 	lifecycle *StreamLifecycle
 	// pendingSig is a thoughtSignature from a text part with no functionCall
 	// sibling, held for the next tool_call chunk.
-	pendingSig string
+	pendingSig      string
+	currentToolID   string
+	currentToolName string
 
 	usageSink UsageSink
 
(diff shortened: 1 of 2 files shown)


Adityakk9031:
Closing PR: Gemini native streaming produces complete, atomic functionCall parts per tool call. Attempting to concatenate function call chunks across events collapses same-name parallel tool calls and produces invalid concatenated JSON.

PR #1092, opened 2026-08-27
translate: strip Anthropic native server tools on non-Anthropic emit

aminsamir45:
## Why

Anthropic executes `web_search_*` / `web_fetch_*` server-side — the model only ever sees the synthesized result blocks. On the Anthropic→non-Anthropic emit paths those declarations were converted like any other tool: `writeOpenAIToolsFromAnthropic` iterates every entry in `tools`, so

```json
{"type":"web_search_20250305","name":"web_search"}
```

(no `input_schema`) became

```json
{"type":"function","function":{"name":"web_search"}}
```

The upstream model can then call a tool the client never registered a handler for, producing a `tool_use` block the harness cannot answer. That is the same phantom-tool failure `claudecode_tool_filter.go` was written for — 224 phantom CC-only `tool_use` blocks on the v0.57 SWE-bench eval.

`websearch.StripServerTools` already did exactly this, was tested, and had **no production caller**.

## What

- Wire `StripServerTools` into `buildOpenAIFromAnthropic`, `buildResponsesFromAnthropic`, and the Gemini `FormatAnthropic` branch, next to the existing Claude-Code-only tool filter, behind a new `RequestMutationStats.ServerToolsStripped`.
- Widen the helper from `web_search_` to a `nativeServerToolPrefixes` list including `web_fetch_`. Writing the tests surfaced this: `web_fetch` survived the strip and became a phantom tool too. `serverToolPrefix` stays search-only so `FindServerTool`/`DetectSearchTurn` keep their narrower meaning — that pair is about the one tool the router knows how to *emulate*, a strictly smaller set than the tools a non-Anthropic upstream cannot *run*.
- Anthropic→Anthropic passthrough is untouched: that upstream is the one that can execute them.

## Behaviour

**None today.** The HMM policy still forces every `web_search`-advertising turn onto an Anthropic arm (`_native_web_search_decision`), so this path is unreachable. This is the groundwork that makes narrowing that guard safe — without the strip, narrowing it would send main-agent turns to OpenAI with a phantom `web_search` function attached.

Context for why the guard is worth narrowing: on the 50-task SWE-Together run `hmm-v65c-swetog-50t-cc2197-20260827b`, the guard fired on 5,750 calls costing **\$738.32 — 66.8% of the run's spend** — and the decision reasons show the policy had chosen `openai/gpt-5.6-luna` for **100%** of them. None were real search turns (20k–58k-token prompts; `DetectSearchTurn` requires a single-message search sub-turn).

## Testing

- New `internal/translate/server_tool_strip_test.go`: OpenAI strips, Gemini strips, Anthropic→Anthropic keeps, and a server-tools-only body drops the now-dangling `tool_choice`.
- `go build ./...`, `go vet`, and the full `go test ./...` suite pass on this base.

### DIFF
--- a/internal/providers/provider.go
+++ b/internal/providers/provider.go
@@ -604,6 +604,13 @@ type RequestMutationStats struct {
 	// CCOnlyToolsStripped counts Claude-Code-only tools removed before
 	// dispatching to a non-Anthropic upstream. See claudecode_tool_filter.go.
 	CCOnlyToolsStripped int
+	// ServerToolsStripped counts Anthropic native server tools (web_search_*,
+	// web_fetch_*) removed before dispatching to a non-Anthropic upstream.
+	// Those tools are executed by Anthropic, not by the model: converted as
+	// ordinary function tools they would become phantom client tools the
+	// caller never registered, so the model can emit a tool_use block the
+	// harness cannot answer. See websearch.StripServerTools.
+	ServerToolsStripped int
 	// GeminiReminderInjected is true when the Gemini 3.x tool-use reminder was
 	// appended to systemInstruction. See translate/system_reminder.go.
 	GeminiReminderInjected bool
--- a/internal/providers/provider.go
+++ b/internal/providers/provider.go
@@ -605,11 +605,9 @@ type RequestMutationStats struct {
 	// dispatching to a non-Anthropic upstream. See claudecode_tool_filter.go.
 	CCOnlyToolsStripped int
 	// ServerToolsStripped counts Anthropic native server tools (web_search_*,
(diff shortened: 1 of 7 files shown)


aminsamir45:
* **#1092** <a href="https://app.graphite.com/github/pr/workweave/router/1092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/workweave/router/1092?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1106, opened 2026-08-28
Improve Codex skill registration

a-churchill:
## Summary

Bring Codex's router-directive UX to parity with Claude Code's.

- Register Codex aliases `$fm`, `$ufm`, and `$rf` alongside the canonical skills.
- **Emit directives from a skill script instead of asking Codex to fake a user message.** A Codex skill is prompt text — it cannot append a `role: user` message or restart the turn. The old skills instructed Codex to "send a normal user message whose first character is one literal space", which has no corresponding capability. In session `01a046f1` the model improvised `text(" /router-feedback …")`, producing a `custom_tool_call_output` that was never parsed as a command; it eventually returned the directive as its final answer. Each prompt skill now ships `scripts/emit.sh`, which Codex execs and which prints the leading-space directive; the router's exec-output parser picks the command out of the tool result.
- Handle agent-issued force-model and router-feedback tool results without terminating the agent turn.
- Sanitize router directives from native Codex Responses passthrough.
- Installer copies and `chmod +x`es the emit scripts; uninstall removes them so a skill dir isn't left without its `SKILL.md`.
- Add installer, packaging, parser, proxy, and regression coverage.

## Verification

- `go test ./...` — pass
- `bash install/tests/registry_test.sh` — 45 passed, 0 failed
- `bash install/tests/packaging_test.sh` — 16 passed, 0 failed
- `bash install/tests/codex_install_test.sh` — pass
- `git diff --check` — clean

Not yet verified against a live Codex CLI session; the emit-script path is covered by the parser/installer tests only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,9 +180,10 @@ matching the Claude Code plugin's model-to-credential dispatch.
 Codex does not load third-party slash-command files, so the installer ships the
 router directives as native Codex skills: `$force-model <model-id>` (alias
 `$fm <model-id>`), `$unforce-model` (alias `$ufm`), and
-`$router-feedback <text>` (alias `$rf <text>`), each of which sends the
-leading-space prompt form (for example, ` /force-model gpt-5.6-terra`) that the
-router parses. You can type that form directly instead. Re-install
+`$router-feedback <text>` (alias `$rf <text>`). Each skill runs a local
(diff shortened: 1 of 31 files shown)


a-churchill:
superseded

PR #1107, opened 2026-08-28
Allow non-Anthropic models to use ToolSearch

a-churchill:
## Summary
- preserve Claude Code `ToolSearch` when translating Anthropic requests to OpenAI and Gemini
- keep other client-only tools filtered, while preserving ToolSearch schemas and accounting
- add deferred-MCP fixture coverage for chat, Responses, Gemini, and passthrough behavior

## Tests
- `go test ./...`


a-churchill:
tested locally!

PR #1120, opened 2026-08-28
Codex directive parity: aliases, toggles, and models

munir-weave:
Brings Codex to parity with the Claude Code directive surface. **Supersedes #1106**, whose commits are cherry-picked here (authorship preserved) and reconciled with the installer refactor below.

## Problem

Codex had 4 of Claude Code's 8 directives, and two of those four were partly broken:

1. **Aliases were advertised but never installed.** `$force-model`, `$unforce-model`, and `$router-feedback` each told the user they could invoke `$fm`/`$ufm`/`$rf`. Codex discovers skills by *directory name*, and no such directories existed — so those aliases could never resolve.
2. **The skills asked Codex to do something it cannot do.** They instructed it to "send a normal user message whose first character is one literal space." A skill is prompt text; it cannot author a `role: user` message. In session `01a046f1` the model improvised `text(" /router-feedback …")`, which became a `custom_tool_call_output` that was never parsed as a command and ended up returned as the final answer.
3. **`router-off`/`on`/`status`/`models` were gated off** for Codex in the registry — not for lack of a mechanism (`$disable-routing` had shipped as a toggle skill for a while) but because the rows were never enabled. No way to check routing state, re-enable it conversationally, or manage the model list.

## Change

- **Prompt skills exec a `scripts/emit.sh`** that prints the leading-space directive; the router's exec-output parser picks it up. Real capability, verified live against Codex 0.150.1 — including the two-part `custom_tool_call_output` array shape the parser actually receives.
- **Aliases install as their own skill directories**, so `$fm`/`$ufm`/`$rf` resolve.
- **Four new toggle skills** — `$router-status`, `$router-off`, `$router-on`, `$router-models` — each shelling out to the same `weave-router <verb> --codex` the CLI exposes, so there's no second copy of the toggle logic.
- **One skill installer.** `install_codex_prompt_skills` walks `weave_registry_skill_assets`; `install_codex_disable_routing_skill` is deleted (it was that loop hardcoded to one directive).
- **`models --codex`** works, resolving the endpoint via the already-per-client `resolve_installed_endpoint`.

### On the `models` trust check

`models_endpoint_is_trusted` guards a Claude-specific hazard: endpoint in committed `settings.json`, key in gitignored `settings.local.json`, so a hostile repo could redirect a teammate's key. Codex keeps both in one managed block — the same-file case that rule already treats as always-safe. Not weakened; Claude's path is byte-identical.

## Testing

`make test-install` green across all six suites (66 / 82 / 50 / 17 + two pass-only). `go test ./...` green. Verified by hand: all 11 skills install and uninstall cleanly, `models --codex` sends the Codex install's key to that install's `/admin/v1/models`, `--claude` is unchanged, and a client with no install errors cleanly.

New test: **every `$name` a skill advertises must resolve to an installed directory.** That is the regression for bug (1) above. I confirmed it fails on a planted fake alias rather than trusting that it passes.

## Bugs found while merging

- Packaging was **failing silently** — `packaging_test.sh` takes the tarball name from `npm pack | tail -1`, and the new `emit.sh` log lines pushed it off the end. Surfaced as an empty test run with exit 0.
- Alias expansion emits names with no skill asset (`models` is a `router-models` alias for the Claude command surface only). The installer already skipped those; packaging threw on the first one. It now skips alias-only names while still failing loudly if a *canonical* declared skill is missing.
- #1106's packaging threw when `scripts/emit.sh` was absent — correct for prompt skills, wrong for the toggles added here, which legitimately ship none.

Earlier review round (Bugbot/Greptile) also caught: `models` needs `jq` (12 call sites) on a Codex-only box that has no reason to have it; the list footer/usage/rejected-key hints hardcoded `--claude`; and a `$models` alias I advertised without installing. All fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,9 @@ sidecar is optional. HMM and forced selections in the native Codex family
 every other selected model uses its WorkWeave deployment or BYOK credential,
 matching the Claude Code plugin's model-to-credential dispatch.
 Codex does not load third-party slash-command files, so the installer ships the
-router directives as native Codex skills: `$force-model <model-id>`,
-`$unforce-model`, and `$router-feedback <text>`, each of which sends the
+router directives as native Codex skills: `$force-model <model-id>` (alias
+`$fm <model-id>`), `$unforce-model` (alias `$ufm`), and
+`$router-feedback <text>` (alias `$rf <text>`), each of which sends the
 leading-space prompt form (for example, ` /force-model gpt-5.6-terra`) that the
 router parses. You can type that form directly instead. Re-install
 and `--uninstall --codex` rewrite/remove only the managed block, leaving the
--- a/README.md
+++ b/README.md
@@ -180,9 +180,10 @@ matching the Claude Code plugin's model-to-credential dispatch.
 Codex does not load third-party slash-command files, so the installer ships the
 router directives as native Codex skills: `$force-model <model-id>` (alias
 `$fm <model-id>`), `$unforce-model` (alias `$ufm`), and
-`$router-feedback <text>` (alias `$rf <text>`), each of which sends the
-leading-space prompt form (for example, ` /force-model gpt-5.6-terra`) that the
-router parses. You can type that form directly instead. Re-install
+`$router-feedback <text>` (alias `$rf <text>`). Each skill runs a local
+`scripts/emit.sh` that prints the leading-space directive (for example,
+` /force-model gpt-5.6-terra`); the router intercepts that exec output.
+You can type that form directly instead. Re-install
 and `--uninstall --codex` rewrite/remove only the managed block, leaving the
 rest of your Codex config untouched. Codex also gets `$router-status`,
 `$router-off`, `$router-on`, and `$router-models` as skills that call this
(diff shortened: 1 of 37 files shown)


munir-weave:
On the High-severity "Custom tool commands stripped unapplied" finding — I checked it and it's a false positive, but a worthwhile one.

`StripRouterCommandsFromResponsesInput` operates on `nativeBody` (the verbatim passthrough bytes). Command extraction runs on the **chat projection**, which `ConvertResponsesToChatCompletionsWithOptions` builds from `conversion.OriginalBody` independently of the strip. The two don't share a buffer, so the directive survives into `ExtractForceModelCommand`/`ExtractRouterFeedbackCommand` while still being sanitized out of what reaches the upstream — which is exactly the intent.

I verified this against the real two-part `custom_tool_call_output` payload captured from a live Codex 0.150.1 session, in `ProxyOpenAIResponses`' exact order (convert, then strip), and added `TestStripRouterCommandsFromResponsesInput_LeavesChatProjectionIntact` to pin the invariant. If someone later aliases those buffers, that test fails instead of every agent-issued $fm/$rf silently disappearing.

The Medium-severity "Preamble drops multi-line feedback" finding on the same commit **was** real and is fixed in b304fead.

PR #1131, opened 2026-08-31
fix(translate): widen the text-only-turn nudge's leaked-markup gate

steventohme:
## Problem

The text-only-turn nudge (`synthesizeTextOnlyTurnNudge`) exists to rescue a turn whose tool call leaked into the content channel as XML — Claude Code's parser rejects that, and the turn dead-ends with no tool call for the client to dispatch. Its gate was:

```go
var toolishMarkupMarkers = []string{"<tool_call", "<function", "<invoke"}
// matched with strings.HasPrefix, anchored to the turn's start
```

A leak does not reliably begin at the call's opening tag. GLM-5.3 streams the opening tags as **reasoning** and only the tail as **content**, so the visible turn opens on a *closing child* tag:

```
</arg_key><arg_value>select:…</arg_value></tool_call>
```

No prefix matched, so the nudge stayed silent and the turn dead-ended.

## Production evidence

Session `b56057cf` dead-ended three consecutive turns — each 31 output tokens, `finish_reason=stop`, **0** tool_use blocks, `text_only_turn_nudged=false`. User nudges ("keep going") could not recover it; the session only resumed when the user forced a different approach ~2 minutes later, and the struggle detector did not escalate for another 10 minutes (it fires on `turn_wall`, which was 605s).

Sampling the affected model's traffic over 3 days: **209 of 400** turns with `has_tools=true, tool_use_blocks=0` landed on `finish=stop / end_turn` with the nudge never firing. Short tool-free dead-end turns run 0.59% on this model and 2.5% on two others; all Anthropic-served models are 0.00%.

## Changes

1. **Closing and child tags added to `toolishMarkupMarkers`** — an anchored match still fires when the turn opens mid-call.
2. **New `containsToolishMarkup`** for leaks that begin with prose and escape the anchored check. It matches only two shapes on purpose: a closing `</tool_call>`, or an **ordered** `<arg_value>…</arg_value>` pair. Prose explaining tool syntax names the *opening* tags, so matching those unanchored would misfire on exactly the prose the anchored check was built to protect. The pair is order-checked so a stray closing tag quoted in prose doesn't qualify.
3. **`leadingContentCap` 64 → 512.** Load-bearing, not cosmetic — the observed leak was 125 bytes, so its closing tags fell *outside* the retained window and no containment check could have seen them. The constant has no other users.

Unchanged: the `requestHadTools && !toolUseEmitted && started` precondition, and every existing suppression (Gemini-3.x sig-less-tool_use guard, `finish_reason=length`, already-suppressed structured calls, clean-prose final answers).

## False-positive check

Replayed the new gate against a week of real production assistant text (443 distinct text blocks, 12k sampled log tails): fires on the **2** leaked-call payloads and **nothing else**.

Also replayed both production payloads byte-for-byte and confirmed the old gate detected neither, so the regression premise holds.

## Tests

5 added (17 total in the nudge suite, all passing):

| Test | Asserts |
|---|---|
| `_FiresOnLeakedCallOpeningMidCall` | the exact production shape now nudges |
| `_FiresOnLeakedCallAfterProse` | leak after leading narration is caught by containment |
| `_LeakBeyondOldWindowStillFires` | closing tags past the old 64-byte cap are still seen |
| `_SkippedWhenProseDiscussesOpeningTagsOnly` | prose naming `<tool_call>`/`<arg_key>` must not nudge |
| `_SkippedWhenProseQuotesUnpairedClosingArgTag` | a lone `</arg_value>` in prose must not nudge |

`go build ./...`, `go vet ./...`, `gofmt -l`, and the full `go test ./...` are clean on top of current `main`.

## Scope

This is the narrow fix: it addresses XML-shaped leaks only. It does **not** help non-XML dead-ends (a model that simply stops emitting tool calls), which would need a shape-independent consecutive-tool-free-turn break. Deliberately left for a separate change.

Note for reviewers: the `text_repetition` detector (#665) could not have caught this either — its window resets at the last *human* turn, so a user typing "keep going" between repetitions structurally disarms it. Also worth a separate look.

### DIFF
--- a/internal/translate/stream.go
+++ b/internal/translate/stream.go
@@ -1220,11 +1220,9 @@ const leadingContentCap = 64
 // the arguments as content, so the visible turn opens mid-call on a closing
 // child tag.
 //
-// Matching stays anchored rather than scanning the whole window: a finished
-// answer that quotes completed markup (`</tool_call>`, a balanced
-// `<arg_value>…</arg_value>`) is indistinguishable from a leak by containment
-// alone, and nudging it appends a fabricated tool call to an already-complete
-// turn. The anchored form catches the observed leaks on its own.
+// Matching stays anchored: quoted completed markup (`</tool_call>`,
+// `<arg_value>…</arg_value>`) is indistinguishable from a leak by
+// containment alone — anchored matching catches both observed leaks.
 //
 // Reasoning markup (<think>) is deliberately excluded: models like Mimo-v2.5
 // stream visible chain-of-thought as <think>…</think> then a real answer with
--- a/internal/translate/stream.go
+++ b/internal/translate/stream.go
@@ -1205,21 +1205,37 @@ func (t *AnthropicSSETranslator) emitValidatedToolArgsDelta(blockIdx int) error 
 // dispatchable; the echo surfaces as a tool_result nudging the model to act.
 const routerNudgeCommand = "echo '[router] previous turn produced no tool_use; use Edit/Write/Read/Bash/Grep — do not respond with prose or thinking tags only.'"
 
-// leadingContentCap bounds retained opening text for leadsWithToolishMarkup —
-// enough for any marker plus leading whitespace.
-const leadingContentCap = 64
+// leadingContentCap bounds retained opening text for leadsWithToolishMarkup
+// and containsToolishMarkup. Sized past any single marker because the
+// containment scan needs the *closing* tag in the buffer too: GLM-5.3 emits
+// `</arg_key><arg_value>…</arg_value></tool_call>` where the payload between
+// the tags is a tool name plus arguments, so the closing tags land well past
+// a marker-width window (the production session's leak was 125 bytes).
+const leadingContentCap = 512
 
-// toolishMarkupMarkers are opening tokens of a tool call leaked into the
-// content channel as XML instead of a structured tool_use block; Claude
-// Code's strict parser rejects these and dead-ends the turn, which the nudge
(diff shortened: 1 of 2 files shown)


steventohme:
Thanks — the P1 is real and I reproduced it locally before changing anything. Driving the translator with a `deepseek-v3.2` tool-enabled stream, both quoted forms produced `nudged=true, tool_use_blocks=1, stop_reason="tool_use"` on an answer that had already completed.

**Removed the fallback rather than narrowing it.** I checked whether it was load-bearing first: the widened *anchored* marker list catches both observed production payloads on its own, so the containment scan bought only a hypothetical case (leak beginning after prose) in exchange for a confirmed false positive that fabricates tool calls on finished turns. Narrowing it further isn't viable either — quoted completed markup is genuinely indistinguishable from a leak by containment, which is the point you made.

Knock-on: `leadingContentCap` reverts 512 → 64. The raise existed *only* so the containment scan had a window wide enough to hold a payload's closing tags; anchored matching needs room for a single marker, and I verified both payloads are still caught at 64.

So the production change is now one line — widen `toolishMarkupMarkers`, keep matching anchored.

Regressions added per your suggestion: quoted closing `</tool_call>` and quoted balanced `<arg_value>…</arg_value>`, each asserting `end_turn` survives with zero `tool_use` blocks and no `toolu_router_nudge_` in the body. I also swapped the now-obsolete cap test for a marker-split-across-SSE-deltas case. 18 nudge tests pass.

Re-ran the false-positive sweep against a week of production assistant text (443 distinct blocks) on the final anchored-only gate: fires on the 2 leaked-call payloads, nothing else.

PR #1153, opened 2026-09-02
feat(router): give the beta HMM strategy its own declarative roster

Symbiomancer:
## Summary

The beta HMM router (#1006) is wired without a Go-side arm selector. Since Go selection became authoritative, a router without a selector negotiates the pre-v3 policy contract, and the HMM sidecar now serves `policy_router_v3` only, so every `/beta` turn would fail closed (503) the moment a beta sidecar is configured. Stable avoids this through `ROUTER_HMM_ROSTER_PATH`, but a beta package embeds its own roster, usually a different cluster taxonomy (the pinned candidate is four-cluster, stable is five), so stable's roster cannot select its arms either.

This adds `ROUTER_HMM_BETA_ROSTER_PATH`: the beta package's roster is loaded with the same `rosterdata.Load` path and the same `selection.Selector` is installed on the beta router. A missing or invalid beta roster disables beta with an error log rather than failing boot, matching the existing treatment of a malformed beta client. Documented in `docs/CONFIGURATION.md` alongside the previously undocumented beta sidecar variables.

## Deployment

WorkWeave sets the variable from terraform (`router_hmm_beta_roster_object`, same `gcs-hmm-roster` mount as stable) in workweave/WorkWeave PR "feat(hmm): serve a v1 or v2 Gemini raw-V5 package through /beta in prod". The roster object for the pinned beta package is already published at `weave_registry/hmm_rosters/roster_96b5c52b18b13f86.json`. Until the gitlink bump lands, the env var is simply unread.

## Verification

- `gofmt -l cmd/router/main.go` clean; `go build ./...`, `go vet ./cmd/router/`.
- `go test ./cmd/router/... ./internal/router/hmm/... ./internal/router/policy/...` pass; full `go test ./...` reported no failing packages.


Symbiomancer:
WorkWeave half (terraform variable, pin, v1/v2 lineage support): workweave/WorkWeave#12861.

PR #1162, opened 2026-09-03
feat(providers): add native MiniMax support

octo-patch:
Reason: Add native MiniMax routing for the current text models with regional OpenAI-compatible endpoints.

- Register MiniMax through the shared OpenAI-compatible client with global and mainland-China endpoint selection.
- Add native catalog bindings with current upstream IDs, pricing, and cache rates while preserving existing route preference.
- Document the MiniMax settings and cover regional URL selection, provider dispatch, catalog resolution, and model rewriting.

Checks:
- `go test ./...`
- `go vet ./internal/providers/... ./internal/router/catalog ./internal/router/rl ./internal/translate ./cmd/router`
- `git diff --check`


octo-patch:
Updated this branch with the latest `main` and resolved the configuration conflict while preserving the MiniMax environment settings. I ran `go test ./...`, `go vet ./internal/providers/... ./internal/router/catalog ./internal/router/rl ./internal/translate ./cmd/router`, and `git diff --check`.

PR #1221, opened 2026-09-08
fix: keep Anthropic tool aliases stable

aminsamir45:
## Summary
- keep generated Anthropic tool aliases stable when sanitized names appear in replayed history
- reserve only live declared tool names when avoiding alias collisions
- add a two-turn regression test covering an alternate alias replay

## Why
Historical tool names are request history, not live declarations. Including them in the alias uniqueness set causes a previously emitted alias to change on the next turn, breaking consistency between the current tool definition and earlier tool calls.

## Validation
- targeted `internal/translate` sanitizer tests
- `wv router typecheck`
- `go vet ./internal/translate`
- `git diff --check`

---
## Summary by cubic
Fixes Anthropic tool alias stability so sanitized tool names keep the same alias across replayed turns.

Previously, historical `tool_use` names in messages were reserved alongside live declarations, which could shift the alias on the next turn. Now only live declared tool names are reserved. Also adds a two-turn regression test covering alternate alias replay.

<sup>Written for commit 06ad42590f73c49776469ce0432c9327a4cc2036. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/weave-os/router/pull/1221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/translate/crossformat_request_test.go
+++ b/internal/translate/crossformat_request_test.go
@@ -2152,29 +2152,46 @@ func TestSanitizeAnthropicHistoricalToolNameCountsUnicodeCharacters(t *testing.T
 	assert.Regexp(t, `^invalid_tool_[a-f0-9]{40}$`, gjson.GetBytes(prep.Body, "messages.0.content.0.name").String())
 }
 
-func TestSanitizeAnthropicToolNamesAvoidsHistoricalAliasCollisions(t *testing.T) {
-	invalidDeclaredName := strings.Repeat("b", 65)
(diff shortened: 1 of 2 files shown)


lowellh315:
nice job on this one samir :)

aminsamir45:
thanks anonymous developer! @lowellh315

PR #1246, opened 2026-09-09
fix: prevent classifier failures from causing router 503s

makosblade:
Superseded by two independently validated PRs, both targeting main:

- #1248 fixes classifier text-budget exhaustion that can erase the user's request at nonzero message positions, with focused regressions and the Go-to-Python contract gate.
- #1249 adds authorized serving recovery for classifier failures, with bounded retries, session/stream correctness, degraded readiness and outage regression coverage.

The compaction root-cause fix is already on main through #1240. The replacement PRs separate the remaining text-budget correction from broader classifier recovery. Neither PR deploys the managed router.


aminsamir45:
I fixed this!

PR #1254, opened 2026-09-09
fix: derive client budgets per turn and constrain compaction recovery

makosblade:
## Summary
- Resolve typed client-budget evidence on each Anthropic request and preview before model normalization or upstream beta injection.
- Keep provider capacity separate from supported harness defaults; report beta-only long-context signals and unobservable private thresholds honestly.
- Add concise client-summary guidance and a bounded initial serial-tool recovery period, preserving history and mapping the tool limit to OpenAI Chat/Responses.
- Handle the real client's trailing-system compaction instruction without rewriting existing messages.

## Validation
- `dots workflow precommit --fix --json`: formatting and Go vet passed.
- `make inference-boundary`: passed.
- `dots workflow test-verify ./... --json`: 5,208 passed, 4 explicit skips, zero failures.
- Weave review criteria: full suite run; naming/reuse fixes applied, unsupported annotation/naming suggestions declined with evidence.
- Installed Claude Code 2.1.257 through the actual proxy and adapter, with a scripted loopback upstream: matched control hit the guard after 3 compactions; recovery completed with 1 compaction and 5 resumed requests. The summary guidance and temporary serial-tool parameter were observed upstream.

## Limits
The canary uses scripted summaries and synthetic usage counters, not live-model compliance or token-accuracy measurements. A single tool result can still exceed the client budget. Private settings/account caps remain unobservable; no new installation headers or live configuration changes are required. No merge or deployment is included.

---
## Summary by cubic
Resolves client context budgets per request instead of assuming provider capacity, and constrains the recovery period after client compaction.

- Computes a Claude Code budget per request (version, model variant, 1M beta) before normalization and upstream beta injection, including provider-qualified and date-suffixed models; route preview resolves the same way from header identity.
- Defers to the client's own compaction only when a supported harness default is known and the pool can serve it; unknown or ambiguous budgets never borrow a provider window.
- Adds a bounded recovery mode that forces serial tool calls for the first resumed turns after client compaction, detected from pre-handover history so it survives handover rewrites, and appends a concise handoff instruction to the summary call.
- Bypasses semantic-cache replies during recovery so older parallel batches can't skip the serial limit.

<sup>Written for commit 310c1e7937004967e6c442224682a73a40c714d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/weave-os/router/pull/1254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/CLIENT_CONTEXT_BUDGET.md
steventohme, line 1: can we remove this i think this should be for more public facing docs not implementation details

steventohme (dismissed):
damn ty

PR #1269, opened 2026-09-10
feat(codex): answer a bare router-models directive server-side

munir-weave:
A bare `$router-models` / `/router-models` (and the `models` alias) is a read, and the router can answer it from the request it is already holding. It now does — the fourth directive to stop costing a model turn, after `force-model`, `router-feedback` and `router-session`.

## The listing got better, not just cheaper

The skill shells out to `weave-router models`, which asks `/admin/v1/models`. On a **managed** router that API isn't mounted at all, so it 404s, degrades to the unauthenticated catalog, and has to print:

> This router does not report which of them your installation has enabled: selection is an organization-wide setting here.

The installation's exclusions are already on the chat request. So the router can just answer the question:

```
anthropic
  [x] claude-opus-5
  [x] claude-haiku-4-5
openai
  [ ] gpt-5.5

2 of 3 routable here. Change the selection with `$router-models enable <id>` …
```

Grouping is by primary catalog binding, with both group order and ids sorted, so the listing doesn't reshuffle between turns on Go's map iteration — there's a test that pins that.

## Only the bare form, and that's the point

`enable` / `disable` / `prefer` / `providers` are admin operations. Those endpoints sit behind `WithAdminOnly`, whose comment says exactly why:

> Mutations: admin cookie REQUIRED. `rk_` tokens are rejected so a leaked data-plane key can't mint fresh router keys or rotate provider credentials.

A chat request carries an `rk_` key. Answering an argument form server-side would be a no-op on managed and **a hole through that boundary** on self-hosted, so anything with arguments falls through to the skill, which authenticates properly. The synthetic footer names that path so the user isn't left guessing.

## Behaviour change worth flagging

Previously a bare `$router-models` started an interactive model turn — it listed models and could take "now turn off gpt-5.5" as a follow-up in the same breath. Now the bare form is a static answer and the change is a second directive. That's the trade for zero inference; I think it's the right one, since the follow-up was never guaranteed to be interpreted the way the user meant, but it is a change.

`directives.tsv` still lists `router-models` as `local-toggle`. That column names the adapter the mutating path needs, and the skill has to keep shipping for it — flipping it to `prompt` would pull it into `weave_registry_skill_names` and change adapter generation. The hybrid is documented in `install/README.md` instead.

## Validation

From the WorkWeave root, per the router guide:

- `wv router tc` — clean
- `wv router t` — green

From `router-internal/router/`:

- `go build ./...`, `go vet ./internal/...`, `gofmt -l` — clean
- `make test-install` — 22 passed, 0 failed
- `make check-docs` — pass

New tests cover all four spellings on ingress, argument forms falling through, the Codex skill-blob case, non-leading uses being ignored, `[x]`/`[ ]` marking against exclusions, per-client footer sigil, the empty-universe case, and ordering determinism.

## Not included

No installer change, so this ships through the gitlink alone — no npm release.

The four genuine local toggles (`$router-on` / `$router-off` / `$router-status` / `$disable-routing`) still cost two turns each; they mutate config on disk, so the router cannot answer them at all. Those need a client-side hook and are a separate change.

---
## Summary by cubic
Answers a bare `$router-models` / `/router-models` (and `models` alias) directive server-side, so it no longer costs a model turn. The listing now marks each model `[x]`/`[ ]`/`[-]` against the installation's restrictions instead of degrading to a flat catalog that can't report what's enabled.

**Behavior change**
- A bare `$router-models` used to accept follow-ups like "now turn off gpt-5.5" in the same turn; it's now a static answer, so that's a second directive.
- A directive with anything else under it is no longer treated as bare: `$router-models\nenable gpt-5.5` falls through to the skill instead of silently dropping the mutation, and `$router-session` with a following question no longer answers the first line and discards the rest.
- Trailing text is only ignored when it's one of the client's own wrappers (`<system-reminder>` and friends); pasted tags like `<error>...</error>` or `<foo>enable gpt-5.5</foo>` now fall through instead of being swallowed.
- `[-]` marks a deployment-wide automatic exclusion, which the scorer won't pick but an explicit `force-model` still serves; the legend only prints when a model is in that state.

**Admin boundary**
- Only bare forms are short-circuited. `enable`/`disable`/`prefer`/`providers` fall through to the skill because they hit admin endpoints that reject the chat request's `rk_` key.

<sup>Written for commit 7155b75695f86a14d5c78aac263632cacdb80407. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/weave-os/router/pull/1269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -310,8 +310,20 @@ router could see the directive behind the skill block. It is retired, and an
 upgrade removes both the hook and its helper.
 
 One group stays skill-driven because it is not router state: the local toggles
-(`$router-on`/`$router-off`/`$router-status`/`$router-models`/
-`$disable-routing`) mutate config on disk, which the router cannot do.
+(`$router-on`/`$router-off`/`$router-status`/`$disable-routing`) mutate config
+on disk, which the router cannot do.
+
+`$router-models` is split. A **bare** invocation is a read, and the router
+answers it directly — with more than the skill could: `weave-router models`
+gets a 404 from the admin API on a managed router and degrades to the plain
+catalog, which is why it has to print "this router does not report which of
+them your installation has enabled". The installation's exclusions are on the
+request, so the router marks each row `[x]`/`[ ]` instead. Anything with
+arguments (`enable`, `disable`, `prefer`, `providers`) still goes to the skill:
+those hit `/admin/v1/*`, which requires an admin session and rejects the `rk_`
+data-plane key a chat request carries — deliberately, so a leaked key cannot
+rewrite routing config. The registry still lists it as `local-toggle` because
+that column names the adapter its mutating path needs.
 
 `$router-session` used to be in that group and no longer is. It reported
 `$CODEX_SESSION_ID` from a script, which cost a model turn plus a tool exec and
--- a/internal/proxy/router_models.go
+++ b/internal/proxy/router_models.go
@@ -30,13 +30,19 @@ func (s *Service) handleRouterModelsCommand(
 	log := observability.FromContext(ctx)
 	routable := s.RoutableModels()
 	excluded := s.excludedModelsForRequest(ctx)
+	// Deployment-wide automatic exclusions are a separate, softer set: the
+	// router may not CHOOSE these, but an explicit force still serves them.
+	// Folding them into `excluded` would claim they are unusable; leaving
+	// them out would claim they are routable. They get their own marker.
+	automaticExcluded := s.globalAutomaticExcludedModels(ctx)
 
 	log.Debug("/router-models: answered from the request's own routing view",
 		"routable", len(routable),
 		"excluded", len(excluded),
+		"automatic_excluded", len(automaticExcluded),
 	)
 
-	msg := routerModelsMessage(routable, excluded,
+	msg := routerModelsMessage(routable, excluded, automaticExcluded,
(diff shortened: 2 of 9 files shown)


munir-weave:
Good catch on the multi-line argument — confirmed and fixed in 8b3b37e.

`parseRouterModelsCommand` matched the token on the first non-empty line and stripped only that line, so `$router-models\nenable gpt-5.5` counted as bare: it answered with a plain listing and dropped the mutation instead of falling through to the skill.

`parseRouterSessionCommand` (already on `main` from #1267) had the same flaw, and worse there — the short-circuit ends the turn, so `$router-session\nand what has this session cost?` answered the first line and discarded the question. Fixed in the same commit.

Both now require the stripped remainder to be empty, via `isOnlyInjectedCommandText` — the predicate the feedback path already uses, so an injected `<system-reminder>` prefix still doesn't count as content. Regression tests added for both.

PR #1270, opened 2026-09-10
feat(codex): answer the four local toggles from a UserPromptSubmit hook

munir-weave:
`$router-on`, `$router-off`, `$router-status` and `$disable-routing` cost **two inference turns each** — one for the model to read the skill and decide to exec, one to report the tool result. They're the last directives paying that, and the only ones the router genuinely cannot answer: they rewrite `config.toml` on the user's machine, which a remote HTTP service can't do.

A `UserPromptSubmit` hook can. It runs the same installer subcommand the skill asked the model to run, and blocks the turn with the output — no inference, and the user sees the command's own text instead of a model's paraphrase of it.

## The turn cost is the smaller half

`$router-off` and `$disable-routing` are what you reach for when the router is slow, broken, or out of credits. As skills they needed a model turn — and that turn was served **through the router**. The off switch depended on the thing it turns off.

The hook runs locally, before any network call. It works when the router doesn't.

## Deliberately narrower than the hook #1257 retired

That one answered `$fm` / `$rf` / `$router-session`, which the router now handles itself. This one claims **only** what the router cannot, so there's no second path to any directive. A test pins it — `$fm`, `$rf`, `$router-session` and `$router-models` must all pass through untouched:

```
fail open
  ok   ordinary prose reaches the model
  ok   router-owned $fm astra is left to the router
  ok   router-owned $rf - too slow is left to the router
  ok   router-owned $router-session is left to the router
  ok   router-owned $router-models is left to the router
  ok   a toggle with an argument passes through
```

## Fail open, everywhere

No `jq`, no `npx`, a malformed payload, an argument on a directive that takes none, a timeout — all hand the prompt to the model untouched. A hook that swallows a prompt is worse than one that never fires.

The one deliberate exception: a **non-zero exit from the toggle itself** is reported, not swallowed. Passing through there would hand the model a prompt the user meant as a command, and it would likely just try the same command again.

Two details carried over from bugs we already hit:
- The blocking response sets `decision`/`reason` only. Adding `continue:false` makes `continue` win and renders a bare "Hook stopped" with the reason discarded — the 0.2.17 bug. There's a test asserting the response carries no `continue` key.
- No GNU `timeout` on macOS, so the wall-clock cap polls the child rather than depending on coreutils.

## Registration is gated on ownership, not existence

`install_codex_directive_script` leaves a user-owned file at the helper path alone. Registering that path anyway would hand someone else's script every prompt, so the config guard tests for our marker, not for the file.

Unlike the status helper, this one **never fails the install** — the skills remain a working fallback, so a stray file costs the hook, not the router. Covered by a test that plants an unowned file and asserts the install still succeeds, doesn't overwrite it, and doesn't register it.

## Standalone installer

`install.sh` embeds the hook as a heredoc for the `curl | sh` path, following `codex-status.sh`, with the same diff-against-canonical drift test — that file's comment notes a stale copy has already shipped once, so the check comes with it.

## Version

Bumped to **0.2.19**. This is an installer change, so it reaches users through an npm release rather than the gitlink.

## Validation

- `make test-install` — every suite, including the new `codex-toggle_test.sh` (69 / 95 / 50 / 22 passed, 0 failed)
- `make check-docs`, `make check-agent-guides` — pass
- shellcheck **v0.10.0**, CI's pinned version, over every changed shell file — clean

Shellcheck caught two things worth noting, both fixed: `{{SCOPE}}` tripping SC1083 in a shell file (the first template token to live in one), and a genuine `A && B || C` gotcha in my own test.

## After merge

```
git tag router-v0.2.19 <merge commit> && git push origin router-v0.2.19
```

Existing installs need `npx @weave-os/router@latest --codex` to pick the hook up, and Codex will prompt once to trust it.

---
## Summary by cubic
Answers `$router-on`, `$router-off`, `$router-status`, and `$disable-routing` from a `UserPromptSubmit` hook—the last directives that cost two inference turns each—and moves the bare `$router-models` read server-side. `$router-off` as a skill needed a model turn served *through* the router, so the off switch depended on the thing it turns off; the hook runs locally before any network call and keeps working when the router doesn't.

The hook claims only these four. `$fm`, `$rf`, `$router-session`, and argument forms of `$router-models` pass through untouched, so there's no second path to any directive as there was with the retired 0.2.17 hook.

**Fail-open behavior**

- No `jq`, no `npx`, malformed payload, an argument on a no-argument toggle, or a timeout hands the prompt to the model untouched; a non-zero exit from a toggle is shown to the user, not swallowed.
- The timeout is honestly bounded: a non-numeric or oversized budget falls back to the default, the child runs in its own process group so a timed-out npx/node is killed, and a toggle exiting 124 is reported, not read as a timeout.
- Registration requires our ownership marker on a regular file. A user-owned file or a symlink at the path—even one pointing at a marked file—is left alone and never registered, and the install skips over it rather than failing.
- The blocking response omits `continue:false` (which discards the reason, the 0.2.17 bug), and the "takes effect next launch" trailer is gated on a zero exit so a failed toggle makes no false claim.

**`$router-models` split**

- A bare `$router-models` (or `/models`/`$models`) is answered from the request's routing view, marking each row `[x]`/`[ ]` from the installation's exclusions—the skill's managed-router path degrades to a flat catalog that cannot report those.
- Argument forms (`enable`, `disable`, `prefer`, `providers`) hit `/admin/v1/*`, which deliberately rejects the chat request's `rk_` key, so they fall through to the skill that authenticates properly.

Version bumped to 0.2.19. After merge, tag `router-v0.2.19`; existing installs re-run `npx @weave-os/router@latest --codex` and trust the new hook when Codex prompts.

<sup>Written for commit c4f5232dd13a521a0a66cce7d254c1a8631c2c3c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/weave-os/router/pull/1270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


munir-weave:
Retargeted onto `feat/router-models-server-side` ([#1269](https://github.com/weave-os/router/pull/1269)) — I had branched this off that work rather than `main`, so the diff was showing both changes. It now shows only the hook. Merge #1269 first and GitHub will retarget this to `main` automatically.

PR #1341, opened 2026-09-15
fix: keep Responses streams alive during advancing reasoning

makosblade:
## Summary
- Count nonempty Responses reasoning deltas and completed reasoning summaries/opaque payloads as stall-watchdog progress in both response translators.
- Keep reasoning separate from answer-output timing and throughput, wiring the callback through both provider adapters and the stream-cut observer.
- Preserve byte-idle protection, the existing request deadline, buffered-client behavior, and the no-replay-after-commit guard. Stall logs no longer promise a retry.

## Validation
- Full Go suite: `go test -vet=all -count=1 ./...`
- Focused translator, provider, watchdog, timing, proxy, and dispatch tests
- Race checks for the new stream/callback cases
- `env -u WEAVE_ROUTER_KEY make precommit` (offline installer tests use synthetic keys)
- `golangci-lint run ./...`: 0 issues
- Weave Checks: all 19 categories pass after targeted fixes/rechecks
- Cumulative review: no remaining correctness or integration findings

Regression coverage includes long reasoning followed by text/tool completion, preserved opaque signatures, first-output/throughput separation, keepalive-only and empty-reasoning stalls, byte silence, cancellation/deadline, and no replay after thinking-only commitment. Fixtures are synthetic; no live upstream calls or deployment changes.


internal/providers/openai/client.go
```diff
@@ -375,18 +375,16 @@ func (c *Client) Proxy(ctx context.Context, decision router.Decision, prep provi
...
-	// Output-progress watchdog (Responses only): the byte-idle watchdog below
-	// resets on ANY byte, so a stream alive on reasoning/keepalive frames but
-	// producing zero output rides to the 600s cap (2026-06-16 incident). This
-	// one measures time-since-last-OUTPUT, fed by the translator only on
-	// output-bearing events, and trips ErrUpstreamOutputStall (retryable).
-	// Wired via ArmOutputProgress since only the translator can tell output
-	// frames from reasoning/keepalive frames.
+	// Translators distinguish advancing reasoning/output from keepalives.
+	// Reasoning feeds only the stall clock, preserving first-output latency.
 	if prep.Endpoint == providers.EndpointResponses {
```
steventohme, line 380: are`/v1/responses` streams are intentionally out of scope? they currently use `responsesTerminalObserver`/ `UsageExtractor`, so reasoning progress won’t reset the watchdog. i think they should be covered too but not sure

steventohme (approved):

internal/providers/openai/client.go
makosblade: ah I think the stall watchdog isn't implement there, that's why. 
Should we add it? I can follow up if so

PR #1353, opened 2026-09-16
fix: coordinate subscription token refresh across replicas

makosblade:
## Problem

Router replicas could concurrently rotate the same pooled subscription refresh token. A stale replica could then disable the shared account and prevent the successful replica from saving its new credentials.

## Change

- Coordinate refreshes with a PostgreSQL lease whose expiry uses the database clock, a fencing UUID, and a refresh-version CAS.
- Encrypt and atomically publish access and refresh tokens so other replicas can adopt the winner’s access token.
- Fence terminal disable and transient cooldown updates on the current lease and credential version, including takeover before the winner publishes.
- Release leases with a bounded context that survives request cancellation; release after adoption without clearing another holder’s lease.
- Keep normal inference on the process-local token cache and surface credential reload failures in diagnostics.
- Use the required sequential migration filename (`0093`) and add deterministic concurrency and cancellation regressions. Synchronize an existing auth test’s log buffer so the auth race suite is reliable.

## Validation

- `make generate`
- `make inference-boundary fmt vet build test`
- `go test -race ./internal/subscriptions ./internal/auth`
- Router precommit checks and migration format/order checks
- `git diff --check`


steventohme (dismissed):
LGTM. ty for this!

Image

aminsamir45:
Lease TTL is 30s and isn’t heartbeaten during refresher.Refresh. If that call outlives the lease, a replica can take over and redeem the same stored refresh token. With single-use rotation the original call may have already invalidated it, so the takeover fails and we can disable/cooldown an account that never got a persisted winner. Can we extend the lease during refresh, or at least not treat that failure as terminal?

aminsamir45:
other than that, LGTM

image

makosblade:
> Lease TTL is 30s and isn’t heartbeaten during refresher.Refresh. If that call outlives the lease, a replica can take over and redeem the same stored refresh token. With single-use rotation the original call may have already invalidated it, so the takeover fails and we can disable/cooldown an account that never got a persisted winner. Can we extend the lease during refresh, or at least not treat that failure as terminal?

Nice catch, will do!

PR #1362, opened 2026-09-17
fix: retry empty completions and raise adaptive Anthropic token floor

makosblade:
## Problem

Some upstream Responses and Chat Completions requests can terminate with a success signal while containing no usable assistant output. The router treated these responses as successful HTTP 200s, leaving clients with an empty turn and no retry. Adaptive reasoning models could also receive the bare 8,192-token default when the caller omitted `max_tokens`, leaving too little budget for hidden reasoning.

## Change

- Add a typed empty-completion error and preserve it through upstream error wrapping so the normal retry path can handle it.
- Reject empty terminal responses across native Responses and Responses-to-Chat/Anthropic translation with a 502 error instead of returning a successful empty response.
- Emit a native `response.failed` terminal event when a streaming response has already committed its headers.
- Apply the existing 16,000-token reasoning floor when injecting defaults for adaptive Anthropic models, while respecting each model's maximum.
- Update proxy fixtures to represent successful upstream responses with actual assistant output.

## Validation

- `go test ./internal/proxy`
- `go test ./internal/translate`
- `go vet ./internal/proxy ./internal/translate`
- All required pull request checks pass, including Go checks, integration tests, Smoke, CodeQL, and review automation.


steventohme (dismissed):
Looks great to me and makes a lot of sense. My agent says this tho, might wanna check it out!

> finalizeEmptyCompletion commits a 502 response and returns nil, so the executor never receives ErrUpstreamEmptyCompletion and cannot retry buffered Responses→Chat/Anthropic requests. This differs from native and streaming paths, which propagate the retryable sentinel. Can we propagate the typed error without committing the response, and add a test proving the buffered path retries?

